Issue #63, opened 2024-08-08
Add v2 endpoints


n8maninger:
- `[GET] /v2/contracts/:id`
-  `[POST] /v2/contracts`
- `[GET] /v2/transactions/:id`
- `[GET /v2/transactions`

Will also need to include the v2 contract resolution

PR #129, opened 2024-10-29
V2 changes

chris124567:
Staging ground for v2 changes

All of the changes merged into here were individually reviewed previously so a review is not strictly required here but if anyone has any final comments that would be fine too.

This implements:
https://github.com/SiaFoundation/explored/issues/63
https://github.com/SiaFoundation/explored/issues/62

### DIFF
--- a/api/server.go
+++ b/api/server.go
@@ -588,7 +588,7 @@ func (s *server) v2ContractsIDRevisionsHandler(jc jape.Context) {
 	jc.Encode(fcs)
 }
 
-func (s *server) pubkeyV2ContractsHandler(jc jape.Context) {
+func (s *server) v2PubkeyContractsHandler(jc jape.Context) {
 	var key types.PublicKey
 	if jc.DecodeParam("key", &key) != nil {
 		return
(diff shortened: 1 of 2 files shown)


chris124567:
All of the changes merged into here were individually reviewed previously so a review is not strictly required here but if anyone has some final comments that would be fine too.

This implements:
https://github.com/SiaFoundation/explored/issues/63
https://github.com/SiaFoundation/explored/issues/62

api/server.go
```diff
@@ -635,8 +701,13 @@ func NewServer(e Explorer, cm ChainManager, s Syncer) http.Handler {
 		"GET    /contracts/:id/revisions": srv.contractsIDRevisionsHandler,
 		"POST   /contracts":               srv.contractsBatchHandler,
 
-		"GET    /pubkey/:key/contracts": srv.pubkeyContractsHandler,
-		"GET    /pubkey/:key/host":      srv.pubkeyHostHandler,
+		"GET    /v2/contracts/:id":           srv.v2ContractsIDHandler,
+		"GET    /v2/contracts/:id/revisions": srv.v2ContractsIDRevisionsHandler,
+		"POST   /v2/contracts":               srv.v2ContractsBatchHandler,
+
+		"GET    /pubkey/:key/v2/contracts": srv.pubkeyV2ContractsHandler,
```
peterjan, line 708: Did we consider making the `key` a filter on whatever route you're interested in? The `v2` here makes it slightly awkward, it kind of looks like the routing scheme is inconsistent, but it is not really. The fewer routes the better and in these handlers you would be calling similar stores as you would in the v2 contracts handler.

chris124567: Do you mean only having one route?

For example we could have:
```
		"GET    /contracts/:id/revisions": srv.contractsIDRevisionsHandler,
```

these routes respond with either v1 or v2 contracts depending on the ID requested. 

But then we wouldn't have a consistent return type for those routes.

n8maninger (approved):

PR #133, opened 2024-10-31
Add v2 file contract revisions

chris124567:


persist/sqlite/v2transactions.go
```diff
@@ -371,6 +352,70 @@ ORDER BY ts.transaction_order ASC`)
...
+	decorateTransactions := func(stmt *stmt, dbID int64, txnIndex int, revision bool) error {
+		rows, err := stmt.Query(dbID)
+		if err != nil {
+			return fmt.Errorf("failed to query file contracts: %w", err)
+		}
+		defer rows.Close()
+
+		j := 0
+		for rows.Next() {
+			fce, err := scanV2FileContract(rows)
```
ChrisSchinnerl, line 389: (nit) if you collect the `fce`s in the helper and then return them all at once instead of assigning them to the txn, you might be able to get rid of the `revision` and `txnIndex` args. Which might clean this up a bit more. Then you would just add 1 line per helper call in the loop which assigns the returned `fces`.

persist/sqlite/v2transactions.go
```diff
@@ -371,6 +352,75 @@ ORDER BY ts.transaction_order ASC`)
...
+			return err
+		}
+
+		revisions, err := collectFileContracts(revisionStmt, dbID)
+		if err != nil {
+			return err
+		}
+
+		if len(parents) != len(revisions) {
+			panic("should have a parent for every v2 revision contract")
```
peterjan, line 410: Is this `panic` absolutely necessary? I can see how we want to, in some cases, panic when we have db corruption but seems quite harsh.

peterjan: I think the statements can be combined using `UNION ALL`, you add something like `"SELECT "parent", ... UNION ALL SELECT "revision"` to differentiate the two and then `collectFileContracts` returns two slices. 

```go
var kind string
if err := rows.Scan(&kind); err != nil {
    return nil, fmt.Errorf("failed to scan contract type: %w", err)
}

fce, err := scanV2FileContract(rows)
if err != nil {
    return nil, fmt.Errorf("failed to scan file contract: %w", err)
}

if kind == "parent" {
    parents = append(contracts, fce)
} else {
    revisions = append(contracts, fce)
}
```

peterjan: Not sure how well this would perform but something this would get rid of the duplicate select expression and clearly show what's happening.

```go
WITH 
    base_data AS (SELECT field1, field2, field3 FROM t1 INNER JOIN t2),
    revision_data AS (SELECT *, "revision" as kind FROM base_data),
    parent_data AS (SELECT *, "parent" as kind FROM base_data)
SELECT * FROM revision_data ORDER BY x
UNION ALL
SELECT * FROM parent_data ORDER BY x
```

peterjan (approved):
LGTM, left some nits and suggestions on `fillV2TransactionFileContractRevisions`

persist/sqlite/v2transactions.go
chris124567: Tried doing something along these lines but I don't think this is worth it unless we find we're having performance issues with the current queries.  It makes the query more complicated and it seems that calling Scan twice on the same Rows doesn't work so we'd need to change the FCE scan helper too because we can't get the `kind` separately from the rest of the data.

PR #136, opened 2024-11-08
Add v2 file contract resolutions

chris124567:
This is the only v2 transaction data we are not presently storing.  I will try to clean up `fillV2TransactionFileContractResolutions` and cut down the number of queries.  We now also store the chain index element for every block which is needed for storage proofs.


persist/sqlite/init.sql
```diff
@@ -327,6 +328,37 @@ CREATE TABLE v2_transaction_file_contract_revisions (
 );
 CREATE INDEX v2_transaction_file_contract_revisions_transaction_id_index ON v2_transaction_file_contract_revisions(transaction_id);
 
+CREATE TABLE v2_transaction_file_contract_resolutions (
+    transaction_id INTEGER REFERENCES v2_transactions(id) ON DELETE CASCADE NOT NULL,
+    transaction_order INTEGER NOT NULL,
+    parent_contract_id INTEGER REFERENCES v2_file_contract_elements(id) ON DELETE CASCADE NOT NULL, -- add an index to all foreign keys
```
peterjan, line 334: Do we have sufficient indices on this table?

explorer/types.go
```diff
@@ -218,6 +217,63 @@ type V2HostAnnouncement struct {
 	chain.V2HostAnnouncement
 }
 
+// V2FileContractResolutionType enumerates the types of file contract resolution.
+type V2FileContractResolutionType interface {
```
peterjan, line 221: Seems like we're not using this interface?

explorer/types.go
```diff
@@ -218,6 +217,63 @@ type V2HostAnnouncement struct {
...
+	isV2FileContractResolution()
+}
+
+func (*V2FileContractFinalization) isV2FileContractResolution() {}
```
ChrisSchinnerl, line 225: Is there a reason for copying all of this from `core`? Duplication should be avoided if possible.

chris124567: In the case of renewals, we use the `V2FileContract` type for the final revision and new contract fields which gives more info than the basic type.  This is not strictly necessary but its in line with what we do everywhere else and it makes things easier for API consumers.  Because the `isV2FileContractResolution` method is unexported, we unfortunately have to copy all the other types (`V2FileContractFinalization`, `V2FileContractExpiration`, `V2StorageProof`) too.

explorer/types.go
chris124567: We need to copy it over from core because of the unexported method

```
impossible type assertion: got.Resolution.(*explorer.V2FileContractRenewal)
        *explorer.V2FileContractRenewal does not implement types.V2FileContractResolutionType (unexported method isV2FileContractResolution
```

persist/sqlite/init.sql
chris124567: I don't think so because we only query this table by transaction_id.

persist/sqlite/v2consensus_test.go
```diff
@@ -3,15 +3,35 @@ package sqlite_test
...
 	"go.sia.tech/explored/explorer"
 	"go.sia.tech/explored/internal/testutil"
+	"golang.org/x/crypto/blake2b"
 )
 
+// from RFC 6962
+const leafHashPrefix = 0
+const nodeHashPrefix = 1
+
+func hashBlockGeneric(msg *[64]byte, prefix uint64) [32]byte {
```
ChrisSchinnerl, line 20: I'd prefer it if we could avoid copying this code over from core and reuse the exposed rhp2 code.

e.g.  the following should work for your test

```go
	data := make([]byte, 2*rhp2.LeafSize)
	data[0], data[rhp2.LeafSize] = 1, 1
	v1FC.FileMerkleRoot, _ = rhp2.ReaderRoot(bytes.NewReader(data))
```

Always better to use the code we also use in production where possible.

persist/sqlite/init.sql
```diff
@@ -9,6 +9,7 @@ CREATE TABLE blocks (
 	parent_id BLOB NOT NULL,
 	nonce BLOB NOT NULL,
 	timestamp INTEGER NOT NULL,
+	chain_index_element BLOB NOT NULL,
```
ChrisSchinnerl, line 12: I don't think you need to store the full element in here since you already got the block id and height. Maybe only the leaf index and proof are missing?

Maybe @lukechampine can verify this assumption.

chris124567:
Should be OK to merge I think.

ChrisSchinnerl (approved):

PR #137, opened 2024-11-08
Store v2 host net addresses

chris124567:
- Store v2 host net addresses in the host_info table
- Fix bug in retrieving transaction events from AddressEvents and add checks
- Fix bug in extracting host announcements from attestations and add checks

Scanning will come in a future PR


persist/sqlite/init.sql
```diff
@@ -444,21 +444,21 @@ CREATE TABLE v2_host_announcements (
...
     event_id INTEGER PRIMARY KEY REFERENCES events(id) ON DELETE CASCADE NOT NULL,
-    transaction_id INTEGER REFERENCES v2_transactions(id) ON DELETE CASCADE NOT NULL,
-    fee BLOB NOT NULL
+    transaction_id INTEGER REFERENCES v2_transactions(id) ON DELETE CASCADE NOT NULL
 );
 
 CREATE TABLE host_info (
     public_key BLOB PRIMARY KEY NOT NULL,
     net_address TEXT NOT NULL,
+    v2_net_addresses BLOB NOT NULL,
```
n8maninger, line 461: Better to separate this into its own table

explorer/types.go
```diff
@@ -299,9 +299,11 @@ type HostScan struct {
 
 // Host represents a host and the information gathered from scanning it.
 type Host struct {
-	PublicKey   types.PublicKey `json:"publicKey"`
-	NetAddress  string          `json:"netAddress"`
-	CountryCode string          `json:"countryCode"`
+	PublicKey      types.PublicKey    `json:"publicKey"`
+	NetAddress     string             `json:"netAddress"`
+	V2NetAddresses []chain.NetAddress `json:"netAddresses"`
```
peterjan, line 304: Fields are a little awkward here. Is there no way to unify them?

ChrisSchinnerl: I don't think there is a good way to do it since v2 addresses include a protocol and because we serve rhp2 over one and rhp4 over the other, but I would make `netAddress` `omit empty` to make sure we drop that field from the response altogether for v2 hosts.

ChrisSchinnerl (approved):

PR #147, opened 2024-12-05
Panic on index error

chris124567:
Panic when we fail to index the data.  This means either there's an erroneous query or the database is out of date and needs to be rescanned anyways.  I leave the logging because panic doesn't give timestamps.


explorer/explorer.go
```diff
@@ -139,6 +139,7 @@ func NewExplorer(cm ChainManager, store Store, batchSize int, scanCfg config.Sca
 			}
 			if err := e.syncStore(lastTip, batchSize); err != nil {
 				e.log.Error("failed to sync store", zap.Error(err))
+				panic(err)
```
n8maninger, line 142: Use log.Panic

n8maninger (changes requested):

n8maninger (approved):

Issue #78, opened 2024-08-30
Use wallet.Event from coreutils


n8maninger:
Replace the `explorer.Event` type with the `wallet.Event` type from coreutils.

PR #149, opened 2024-12-09
Use coreutils/wallet.Event

chris124567:
~~Use the standard event type used in our other apps.  Unfortunately to avoid duplicating queries this requires some conversions from explorer types to core types.  Alternatively, we could just use the enhanced explorer types for the transaction and resolution events which would make things somewhat easier.~~

Use standard event types from used in other apps, but with explorer type versions of the fields rather than the core versions.

Fix https://github.com/SiaFoundation/explored/issues/78

### DIFF
--- a/explorer/events.go
+++ b/explorer/events.go
@@ -19,10 +19,6 @@ type (
 	// siafund elements.
 	EventV1Transaction struct {
 		Transaction Transaction `json:"transaction"`
-		// v1 siacoin inputs do not describe the value of the spent utxo
-		SpentSiacoinElements []SiacoinOutput `json:"spentSiacoinElements,omitempty"`
-		// v1 siafund inputs do not describe the value of the spent utxo
-		SpentSiafundElements []SiacoinOutput `json:"spentSiafundElements,omitempty"`
 	}
 
 	// An EventV1ContractResolution represents a file contract payout from a v1
@@ -122,7 +118,6 @@ func AppliedEvents(cs consensus.State, b types.Block, cu ChainUpdate) (events []
 				continue
 			}
 
-			// e.SpentSiacoinElements = append(e.SpentSiacoinElements, sce)
 			addresses[sce.SiacoinOutput.Address] = struct{}{}
 		}
 		for _, sco := range txn.SiacoinOutputs {
@@ -135,7 +130,6 @@ func AppliedEvents(cs consensus.State, b types.Block, cu ChainUpdate) (events []
 				continue
 			}
 
-			// e.SpentSiafundElements = append(e.SpentSiafundElements, sfe)
 			addresses[sfe.SiafundOutput.Address] = struct{}{}
 
 			sce, ok := sces[sfi.ParentID.ClaimOutputID()]
--- a/explorer/events.go
+++ b/explorer/events.go
@@ -275,9 +275,8 @@ func AppliedEvents(cs consensus.State, b types.Block, cu ChainUpdate) (events []
 			panic("unknown resolution type")
 		}
 
-		efc := V2FileContract{V2FileContractElement: fce}
-		{
-			element := sces[types.FileContractID(fce.ID).V2HostOutputID()]
+		addV2Resolution := func(element types.SiacoinElement) {
+			efc := V2FileContract{V2FileContractElement: fce}
 			addEvent(types.Hash256(element.ID), element.MaturityHeight, wallet.EventTypeV2ContractResolution, EventV2ContractResolution{
 				Resolution: V2FileContractResolution{
 					Parent:     efc,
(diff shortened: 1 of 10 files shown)


persist/sqlite/consensus.go
```diff
@@ -643,84 +643,53 @@ func addEvents(tx *txn, scDBIds map[types.SiacoinOutputID]int64, fcDBIds map[exp
...
-	contractPayoutEventStmt, err := tx.Prepare(`INSERT INTO contract_payout_events (event_id, output_id, contract_id, missed) VALUES (?, ?, ?, ?)`)
+	v1ContractResolutionEventStmt, err := tx.Prepare(`INSERT INTO v1_contract_resolution_events (event_id, output_id, parent_id, missed) VALUES (?, ?, ?, ?)`)
 	if err != nil {
-		return fmt.Errorf("failed to prepare contract payout event statement: %w", err)
+		return fmt.Errorf("failed to prepare v1 contract resolution event statement: %w", err)
 	}
-	defer contractPayoutEventStmt.Close()
+	defer v1ContractResolutionEventStmt.Close()
 
-	foundationSubsidyEventStmt, err := tx.Prepare(`INSERT INTO foundation_subsidy_events (event_id, output_id) VALUES (?, ?)`)
```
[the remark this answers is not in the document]
chris124567: They are classified as `wallet.EventPayout` events.

persist/sqlite/events.go
```diff
@@ -0,0 +1,273 @@
...
+		txn.ArbitraryData = append(txn.ArbitraryData, arb)
+	}
+	for _, sig := range e.Signatures {
+		txn.Signatures = append(txn.Signatures, sig)
+	}
+
+	return
+}
+
+func explorerToEventV2Transaction(e explorer.V2Transaction) (txn wallet.EventV2Transaction) {
```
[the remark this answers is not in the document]
chris124567: The problem is we use many explorer "enhanced" types

explorer/types.go:
```
type SiacoinOutput struct {
    Source     Source            `json:"source"`
    SpentIndex *types.ChainIndex `json:"spentIndex"`
    types.SiacoinElement
}

type V2Transaction struct {
    ...
    SiacoinOutputs []SiacoinOutput        `json:"siacoinOutputs,omitempty"`
    ...
}
```

So if we were to store a types.V2Transaction too it'd basically be a duplicate of a bunch of data

n8maninger:
This is pretty rough. I'm leaning towards using the extended types. They are close enough matches that it will probably be fine.

chris124567:
> This is pretty rough. I'm leaning towards using the extended types. They are close enough matches that it will probably be fine.

:+1: With you on that.  I guess we can still use wallet.Event but the implementing types will be different.

chris124567:
Started on this but it has me thinking.  When we do unconfirmed events, aren't we going to have to convert from the core types into the explorer types?  Because we're going to be taking transactions that haven't been indexed yet more or less straight from the txpool.  Except we might not have all the info to fill out the "enhanced" fields for the explorer types.

explorer/events.go
```diff
@@ -131,144 +95,234 @@ func AppliedEvents(cs consensus.State, b types.Block, cu ChainUpdate) []Event {
...
 		}
+
 		for _, sfi := range txn.SiafundInputs {
-			addrs = append(addrs, sfi.Parent.SiafundOutput.Address)
+			sfe, ok := sfes[sfi.ParentID]
+			if !ok {
+				continue
+			}
+
+			// e.SpentSiafundElements = append(e.SpentSiafundElements, sfe)
```
ChrisSchinnerl, line 138: commented out line of code

explorer/events.go
```diff
@@ -131,144 +95,234 @@ func AppliedEvents(cs consensus.State, b types.Block, cu ChainUpdate) []Event {
...
+					SiacoinElement: SiacoinOutput{SiacoinElement: sce},
+				}, []types.Address{sfi.ClaimAddress})
 			}
 		}
+		for _, sco := range txn.SiafundOutputs {
+			addresses[sco.Address] = struct{}{}
+		}
 
-		addEvent(types.Hash256(txn.ID()), cs.Index.Height, &e, relevant) // transaction maturity height is the current block height
+		// ev := EventV2Transaction(txn)
```
ChrisSchinnerl, line 198: commented out line of code

explorer/events.go
```diff
@@ -131,144 +95,234 @@ func AppliedEvents(cs consensus.State, b types.Block, cu ChainUpdate) []Event {
...
+			addEvent(types.Hash256(element.ID), element.MaturityHeight, wallet.EventTypeV2ContractResolution, EventV2ContractResolution{
+				Resolution: V2FileContractResolution{
+					Parent:     efc,
+					Type:       typ,
+					Resolution: res,
+				},
+				SiacoinElement: SiacoinOutput{SiacoinElement: element},
+				Missed:         missed,
+			}, []types.Address{fce.V2FileContract.RenterOutput.Address})
+		}
```
ChrisSchinnerl, line 309: We usually don't use scopes like this so I'd avoid them for amore consistent code style across our projects.
e.g. in this case an alternative that provides similar safety with less copy-paste risk would be 

```go
		addV2Resolution := func(element types.SiacoinElement) {
			efc := V2FileContract{V2FileContractElement: fce}
			addEvent(types.Hash256(element.ID), element.MaturityHeight, wallet.EventTypeV2ContractResolution, EventV2ContractResolution{
				Resolution: V2FileContractResolution{
					Parent:     efc,
					Type:       typ,
					Resolution: res,
				},
				SiacoinElement: SiacoinOutput{SiacoinElement: element},
				Missed:         missed,
			}, []types.Address{fce.V2FileContract.HostOutput.Address})
		}
		addV2Resolution(sces[types.FileContractID(fce.ID).V2RenterOutputID()])
		addV2Resolution(sces[types.FileContractID(fce.ID).V2HostOutputID()])
```

ChrisSchinnerl (dismissed):

chris124567:
Rebased

ChrisSchinnerl (approved):

PR #152, opened 2024-12-17
Add host filtering endpoint

chris124567:
Add host filtering endpoint similar to https://web.archive.org/web/20240106075954/https://siacentral.com/api/docs/#list-get in Siacentral.  Also support querying the v2/rhp4 versions of fields like contract price if requested by the user.


api/client.go
```diff
@@ -264,3 +264,9 @@ func (c *Client) Search(id types.Hash256) (resp explorer.SearchType, err error)
 	err = c.c.GET(fmt.Sprintf("/search/%s", id.String()), &resp)
 	return
 }
+
+// HostsList searches the hosts by the given criteria.
+func (c *Client) HostsList(params explorer.HostQuery, sortBy explorer.HostSortColumn, dir explorer.HostSortDir, offset, limit uint64) (resp []explorer.Host, err error) {
+	err = c.c.POST(fmt.Sprintf("/hosts/list?sort=%s&dir=%s&offset=%d&limit=%d", sortBy, dir, offset, limit), params, &resp)
```
peterjan, line 270: we generally use `url.Values`, you get some automated encoding but also it increases readability

api/server.go
```diff
@@ -628,6 +629,31 @@ func (s *server) pubkeyHostHandler(jc jape.Context) {
 	jc.Encode(hosts[0])
 }
 
+func (s *server) hostsListHandler(jc jape.Context) {
+	var params explorer.HostQuery
+	if jc.Decode(&params) != nil {
+		return
+	}
+
+	limit := uint64(100)
```
peterjan, line 638: Just wanted to briefly draw attention to the default limit here. In `renterd` we default to unlimited (-1), in `hostd` we default to 100 but apply a 500 max limit. Would be nice to get these small things consistent across our APIs.

api/server.go
```diff
@@ -628,6 +629,31 @@ func (s *server) pubkeyHostHandler(jc jape.Context) {
...
+	offset := uint64(0)
+	if jc.DecodeForm("limit", &limit) != nil || jc.DecodeForm("offset", &offset) != nil {
+		return
+	}
+
+	dir := explorer.HostSortAsc
```
peterjan, line 644: `nit` would reverse the order and return the newer hosts first?

persist/sqlite/hosts.go
```diff
@@ -0,0 +1,315 @@
...
+			sortColumn = "settings_total_storage - settings_remaining_storage"
+		}
+	case explorer.HostSortTotalStorage:
+		if params.V2 {
+			sortColumn = "rhp4_settings_total_storage"
+		} else {
+			sortColumn = "settings_total_storage"
+		}
+	default:
+		panic("invalid sort column")
```
peterjan, line 251: Naked panic? I haven't seen any validation in the header. This should eventually end up returning a 400 status code. I don't mind the store panicking that much but if we do that we should validate it higher up, ensuring that this panic is essentially a developer error.

persist/sqlite/scan_test.go
```diff
@@ -302,4 +303,123 @@ func TestScan(t *testing.T) {
 			log.Fatal("SectorSize = 0 on host that's supposed to be active")
 		}
 	}
+
+	{
+		dbHosts, err := e.QueryHosts(explorer.HostQuery{V2: true}, explorer.HostSortAcceptingContracts, explorer.HostSortAsc, 0, math.MaxInt64)
```
peterjan, line 308: I would extract `QueryHosts` into its own tests and extend it using a table test that verifies every query parameter. We have a lot of sort columns and some of these deal with currencies, so especially for these we have to assert we're performing a numerical sort.

persist/sqlite/hosts.go
ChrisSchinnerl: Honestly I'd just return an error here. No need to ever risk crashing the explorer over this.

persist/sqlite/hosts.go
```diff
@@ -0,0 +1,315 @@
...
+			result = append(result, host)
+		}
+		return nil
+	})
+	return
+}
+
+// QueryHosts returns the hosts matching the query parameters in the order
+// specified by dir.
+func (st *Store) QueryHosts(params explorer.HostQuery, sortBy explorer.HostSortColumn, dir explorer.HostSortDir, offset, limit uint64) (result []explorer.Host, err error) {
```
ChrisSchinnerl, line 123: Why not merge this with the existing `Hosts` endpoint? You already filter by potentially a lot of fields here so you could also filter by public keys and avoid duplicating some of the code.

api/server.go
ChrisSchinnerl: Might be fine for the explorer to be a bit stricter since unknown third parties send these requests where in other places we know only trusted parties have access.

api/server.go
ChrisSchinnerl: I think this is fine. Imo ascending is the more natural default for most fields we could sort by here. Also means the results for lower offsets on the default query don't change as much since newer hosts come last.

api/server.go
```diff
@@ -709,6 +750,8 @@ func NewServer(e Explorer, cm ChainManager, s Syncer) http.Handler {
 		"GET    /metrics/block/:id": srv.blocksMetricsIDHandler,
 		"GET    /metrics/host":      srv.hostMetricsHandler,
 
+		"POST   /hosts/list": srv.hostsListHandler,
```
n8maninger, line 753: `/hosts` is preferable.

persist/sqlite/hosts.go
chris124567: Is there a use case where people want to filter to get a subset of the pubkeys provided versus just a subset of all hosts?  Also, the filtering is v1/v2 different for v1/v2 hosts, so the problem is that someone could have to make two requests to get a host if they are care about hosts with a certain pubkey: one where `params.V2` is false and one where `params.V2` is true.

explorer/types.go
```diff
@@ -368,3 +370,122 @@ type HostMetrics struct {
 	Settings   rhpv2.HostSettings   `json:"settings"`
 	PriceTable rhpv3.HostPriceTable `json:"priceTable"`
 }
+
+// HostSortDir represents the sorting direction for host filtering.
+type HostSortDir string
+
+const (
+	// HostSortAsc means sorting in ascending order.
+	HostSortAsc HostSortDir = "ASC"
```
peterjan, line 379: `nit` I'd use lower case values for sort dir here because they're passed as query string params

persist/sqlite/hosts.go
peterjan: What if we make the V2 param optional, so it can either be unset, true or false. Merging these was the first thing that came to mind when I saw [this](https://github.com/SiaFoundation/explored/pull/152/files#diff-a5457320f6b67f92cbcb51459dd5ebd160912638c6b22d0964e52921b58a05d9R74).

peterjan: Oh I see the problem now. You can solve it but it requires a little rewrite... you essentially have to change all your `if params.V2 then rhp4_setting_x = y else setting_x=y` to `CASE WHEN v2=1 THEN ... ELSE .. END`. That will clean up building the filters and allow you to return both v1 and v2 hosts.

chris124567:
That was helpful PJ.  V2 along with AcceptContracts are now pointers and optional to specify.

peterjan (dismissed):

ChrisSchinnerl (approved):

peterjan (approved):

PR #153, opened 2024-12-21
Fix host scanning bug

chris124567:
Fix scanning bug that has affected the mainnet explored node.  We were not incrementing the offset passed to `HostsForScanning` in `fetchHosts`.  On testnets and in tests this wasn't a problem because the number of hosts was fewer than the batch size (`scanBatchSize`), so incrementing the offset wasn't necessary cause it'd only take one loop iteration in `fetchHosts`.  On mainnet this was a problem though so we'd keep scanning the same hosts over and over and never closing the hosts channel so the scans that were completed were never added to the DB.

### DIFF
--- a/explorer/explorer.go
+++ b/explorer/explorer.go
@@ -67,7 +67,7 @@ type Store interface {
 	SiafundElements(ids []types.SiafundOutputID) (result []SiafundOutput, err error)
 
 	Hosts(pks []types.PublicKey) ([]Host, error)
-	HostsForScanning(maxLastScan, minLastAnnouncement time.Time, offset, limit uint64) ([]Host, error)
+	HostsForScanning(maxLastScan, minLastAnnouncement time.Time, limit uint64) ([]Host, error)
 }
 
 // Explorer implements a Sia explorer.
--- a/explorer/scan.go
+++ b/explorer/scan.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"net"
-	"sync"
 	"time"
 
 	crhpv2 "go.sia.tech/core/rhp/v2"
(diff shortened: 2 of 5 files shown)


n8maninger (changes requested):
You can simplify the scanning logic and reduce the surface area for bugs: combine the fetch and scan goroutines and remove the offset parameter from `HostsForScanning`. Focus on the core logic and keep the complexity as low as possible to reduce the chance for bugs like this to sneak in and make it easier to test. Add a test to the store impl to ensure `HostsForScanning` is returning the expected results. 

This should be about as complex as necessary:
```go
locator, err := geoip.NewIP2LocationLocator("")
if err != nil {
	return fmt.Errorf("failed to create geoip database: %w", err)
}
defer locator.Close()

var wg sync.WaitGroup

for {
	select {
	case <-ctx.Done():
		return ctx.Err() // shutdown
	default:
	}
	
	lastScanCutoff := time.Now().Add(-e.scanCfg.MaxLastScan)
	lastAnnouncementCutoff := time.Now().Add(-e.scanCfg.MinLastAnnouncement)

	// no need for offset parameter since results are added to the database in the same goroutine
	batch, err := e.s.HostsForScanning(lastScanCutoff, lastAnnouncementCutoff, scanBatchSize)
	if err != nil {
		return fmt.Errorf("failed to get hosts for scanning: %w", err)
	} else if len(batch) == 0 {
		select {
		case <-ctx.Done():
			return ctx.Err() // shutdown
		case <-time.After(30 * time.Second):
			continue // check again
		}
	}

	results := make([]HostScan, len(batch))
	for i, host := range batch {
		wg.Add(1)
		go func(i int, host Host) {
			defer wg.Done()

			scan, err := e.scanV1Host(locator, host)
			if err != nil {
				e.log.Debug("host scan failed", zap.Stringer("pk", host.PublicKey), zap.Error(err))
				results[i] = HostScan{
					PublicKey: host.PublicKey,
					Success:   false,
					Timestamp: types.CurrentTimestamp(),
				}
				return
			}
			results[i] = scan
		}(i, host)
	}
	wg.Wait()

	if err := e.s.AddHostScans(results); err != nil {
		return fmt.Errorf("failed to add host scans to DB: %w", err)
	}
}
```

Out of scope for this PR, but it's not idiomatic to store a context on a struct: https://go.dev/blog/context-and-structs

n8maninger (approved):

PR #156, opened 2025-01-09
Add pending events

chris124567:
todo: Fill out parent info in transactions in the function that annotates unconfirmed transactions and add tests

### DIFF
--- a/api/server.go
+++ b/api/server.go
@@ -62,6 +62,7 @@ type (
 		UnspentSiacoinOutputs(address types.Address, offset, limit uint64) ([]explorer.SiacoinOutput, error)
 		UnspentSiafundOutputs(address types.Address, offset, limit uint64) ([]explorer.SiafundOutput, error)
 		AddressEvents(address types.Address, offset, limit uint64) (events []explorer.Event, err error)
+		AddressUnconfirmedEvents(address types.Address) ([]explorer.Event, error)
 		Contracts(ids []types.FileContractID) (result []explorer.ExtendedFileContract, err error)
 		ContractsKey(key types.PublicKey) (result []explorer.ExtendedFileContract, err error)
 		ContractRevisions(id types.FileContractID) (result []explorer.ExtendedFileContract, err error)
@@ -468,6 +469,20 @@ func (s *server) addressessAddressEventsHandler(jc jape.Context) {
 	jc.Encode(events)
 }
 
+func (s *server) addressessAddressEventsUnconfirmedHandler(jc jape.Context) {
+	var address types.Address
+	if jc.DecodeParam("address", &address) != nil {
+		return
+	}
+
+	events, err := s.e.AddressUnconfirmedEvents(address)
+	if jc.Check("failed to get unconfirmed address events", err) != nil {
+		return
+	}
+
+	jc.Encode(events)
+}
+
 func (s *server) outputsSiacoinHandler(jc jape.Context) {
 	var id types.SiacoinOutputID
 	if jc.DecodeParam("id", &id) != nil {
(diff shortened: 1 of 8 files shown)


api/server.go
```diff
@@ -468,6 +474,62 @@ func (s *server) addressessAddressEventsHandler(jc jape.Context) {
...
+		return
+	} else if len(events) > 0 {
+		jc.Encode(events[0])
+		return
+	}
+
+	{
+		v1, v2 := s.cm.PoolTransactions(), s.cm.V2PoolTransactions()
+
+		relevantV1 := v1[:0]
```
peterjan, line 508: these loops are redundant, it looks like you wanted to filter at one point but removed it maybe?

api/server.go
```diff
@@ -468,6 +474,62 @@ func (s *server) addressessAddressEventsHandler(jc jape.Context) {
...
+
+	events, err := s.e.Events([]types.Hash256{id})
+	if err != nil {
+		return
+	} else if len(events) > 0 {
+		jc.Encode(events[0])
+		return
+	}
+
+	{
```
peterjan, line 505: why the explicit scope? seems unnecessary

explorer/events.go
```diff
@@ -65,6 +68,56 @@ func (EventV1ContractResolution) isEvent() bool { return true }
 func (EventV2Transaction) isEvent() bool        { return true }
 func (EventV2ContractResolution) isEvent() bool { return true }
 
+// UnmarshalJSON implements the json.Unmarshaler interface.
+func (e *Event) UnmarshalJSON(b []byte) error {
+	var je struct {
```
peterjan, line 73: I think you missed `confirmations`, maybe we should consider adding a unit test that marshals/unmarshals an event since it's quite an odd type

explorer/events.go
```diff
@@ -73,6 +126,279 @@ type ChainUpdate interface {
 	ForEachV2FileContractElement(func(fce types.V2FileContractElement, created bool, rev *types.V2FileContractElement, res types.V2FileContractResolutionType))
 }
 
+// RelevantAddressesV1 returns all the relevant addresses to a V1 transaction.
+func RelevantAddressesV1(txn types.Transaction) []types.Address {
```
peterjan, line 130: `nit` we're being inconsistent with `V1/V2` suffices, sometimes it's fooV1/fooV2 and other times it's foo/fooV2 - feel free to resolve this comment though, just something to keep in the back of your head, I've noticed it a couple of times but never mentioned it, not only in `explored` but other repos as well

explorer/events.go
```diff
@@ -73,6 +126,279 @@ type ChainUpdate interface {
...
+	}
+
+	relevant := make([]types.Address, 0, len(addresses))
+	for addr := range addresses {
+		relevant = append(relevant, addr)
+	}
+	return relevant
+}
+
+// RelevantAddressesV2 returns all the relevant addresses to a V1 transaction.
```
peterjan, line 168:
```suggestion
// RelevantAddressesV2 returns all the relevant addresses to a V2 transaction.
```

explorer/events_test.go
```diff
@@ -0,0 +1,116 @@
...
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		var unmarshalled explorer.Event
+		if err := json.Unmarshal(data, &unmarshalled); err != nil {
+			t.Fatal(err)
+		}
+
+		testutil.Equal(t, "ID", event.ID, unmarshalled.ID)
```
peterjan, line 107: `nit` perhaps `reflect.DeepEqual` is a more robust choice here

peterjan (approved):

ChrisSchinnerl (approved):

PR #158, opened 2025-01-12
Add exchange rates

chris124567:
- Create interface for exchange rate APIs
- Add CoinGecko and Kraken implementors
- Add implementor that can average exchange rates from different sources
- Add API endpoint for exchange rate in USD

The implementors just retrieve the exchange rate every x seconds and cache the results.  This way the number of requests doesn't scale with the number of users and we won't exceed rate limits.


cmd/explored/main.go
```diff
@@ -25,6 +25,7 @@ import (
 	"go.sia.tech/explored/build"
 	"go.sia.tech/explored/config"
 	"go.sia.tech/explored/explorer"
+	"go.sia.tech/explored/internal/exchangerates"
```
n8maninger, line 28: If it’s used in main, it can’t be in internal

n8maninger (changes requested):
This is a good start, but really should support more than just USD.

chris124567:
This seems pretty extensible.  We could also respond with the rates for all the currencies at once if you think that'd be better.

n8maninger:
I think it's fine for our API to return a single rate per request. However, a timer per source per pair is less than ideal. For example, with Coingecko, you can get all the currencies you need at once: 
```
https://api.coingecko.com/api/v3/simple/price?ids=bitcoin%2Csiacoin%2Csiaprime-coin%2Cbitcoin&vs_currencies=btc%2Cusd%2Ccad%2Ceur%2Cgbp%2Cjpy%2Caus%2Crub%2Ccny%2Ceth&include_market_cap=false&include_24hr_vol=false&include_24hr_change=false&include_last_updated_at=false
```

api/server.go
```diff
@@ -739,12 +741,37 @@ func (s *server) searchIDHandler(jc jape.Context) {
...
+}
+
 // NewServer returns an HTTP handler that serves the explored API.
-func NewServer(e Explorer, cm ChainManager, s Syncer) http.Handler {
+func NewServer(e Explorer, cm ChainManager, s Syncer, exs map[string]exchangerates.ExchangeRateSource) http.Handler {
 	srv := server{
 		cm:        cm,
 		e:         e,
 		s:         s,
+		exs:       exs,
```
n8maninger, line 774: Create an interface and struct for managing exchange rates
```
type Rates interface {
  Last() (types.Currency, error)
}
```

exchangerates/exchangerates.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+type averager struct {
+	ignoreOnError bool
+	sources       []ExchangeRateSource
+}
+
+// NewAverager returns an exchange rate source that averages multiple exchange
+// rates.
+func NewAverager(ignoreOnError bool, sources ...ExchangeRateSource) ExchangeRateSource {
```
ChrisSchinnerl, line 21: I'd sanity check that at least one source is provided

exchangerates/exchangerates.go
```diff
@@ -0,0 +1,51 @@
...
+	for i := range a.sources {
+		go a.sources[i].Start(ctx)
+	}
+}
+
+// Last implements ExchangeRateSource.
+func (a *averager) Last() (float64, error) {
+	sum, count := 0.0, 0.0
+	for i := range a.sources {
+		if v, err := a.sources[i].Last(); err == nil {
```
ChrisSchinnerl, line 39: We might be missing an edge case here. From the looks of it a source might return `0, nil` when calling `Last` right after creating it. So we need to make sure to avoid that. 

One good precaution would be to ignore `v` when it's `0` cause that should never be a valid result anyway.

exchangerates/kraken.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+func newKrakenAPI() *krakenAPI {
+	return &krakenAPI{}
+}
+
+// See https://docs.kraken.com/api/docs/rest-api/get-ticker-information
+func (k *krakenAPI) ticker(pair string) (float64, error) {
+	pair = strings.ToUpper(pair)
+	response, err := k.client.Get("https://api.kraken.com/0/public/Ticker?pair=" + url.PathEscape(pair))
```
ChrisSchinnerl, line 48: Would be a bit cleaner to interrupt the http requests with a context on shutdown that you pass in

exchangerates/kraken.go
```diff
@@ -0,0 +1,116 @@
...
+// Start implements ExchangeRateSource.
+func (k *kraken) Start(ctx context.Context) {
+	ticker := time.NewTicker(k.refresh)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+			k.mu.Lock()
+			k.rate, k.err = k.client.ticker(k.pair)
```
ChrisSchinnerl, line 99: I'd immediately execute `ticker` with `ctx` as its first argument to interrupt the request and afterwards call `select` . That reduces the time before the first update of the value on startup.

ChrisSchinnerl (approved):

exchangerates/exchangerates.go
```diff
@@ -0,0 +1,77 @@
...
+	// CurrencyCNY represents Chinese yuan.
+	CurrencyCNY = "CNY"
+	// CurrencyBTC represents Bitcoin.
+	CurrencyBTC = "BTC"
+	// CurrencyETH represents Ethereum.
+	CurrencyETH = "ETH"
+)
+
+// An ExchangeRateSource returns the price of 1 unit of an asset in USD.
+type ExchangeRateSource interface {
```
n8maninger, line 30: Should be `Source` since it's in the `exchangerate` package

n8maninger (approved):

PR #160, opened 2025-01-17
Exponential delay if scan fails

chris124567:
Add exponential backoff when scanning hosts as described in https://github.com/SiaFoundation/explored/pull/57#discussion_r1746084941.  Also this makes it more justifiable to bump up MinLastAnnouncement to 365 days (like [renterd does](https://github.com/SiaFoundation/renterd/blob/master/cmd/renterd/config.go#L86)) because we will waste much less time scanning hosts that have long been down.

If a scan succeeds, we add MaxLastScan (default 3 hours) to the next scan time.  If it fails, we add MaxLastScan * 2^(number of failures since last success) to the next scan time.  So on the first failure we'd add 6 hours, next failure 12 hours, 24 hours, ....  If we eventually successfully scan that host but then it subsequently fails again, we'd start over and only add 6 hours, if it failed again 12, and so on.


persist/sqlite/hosts.go
```diff
@@ -9,12 +9,14 @@ import (
...
 )
 
-// HostsForScanning returns hosts ordered by the transaction they were created in.
+// HostsForScanning returns hosts ordered by their time to next scan.  Hosts
+// which are repeatedly offline will face an exponentially growing next scan
+// time to avoid wasting resources.
 // Note that only the PublicKey, V2, NetAddress, and V2NetAddresses fields are
 // populated.
-func (s *Store) HostsForScanning(maxLastScan, minLastAnnouncement time.Time, limit uint64) (result []explorer.Host, err error) {
+func (s *Store) HostsForScanning(now, minLastAnnouncement time.Time, limit uint64) (result []explorer.Host, err error) {
```
n8maninger, line 17: There's really no benefit to passing the current time in. Go ahead and use `time.Now()` in the SQLite layer

persist/sqlite/consensus.go
```diff
@@ -1177,9 +1177,9 @@ func addHostScans(tx *txn, scans []explorer.HostScan) error {
 }
 
 // AddHostScans implements explorer.Store
-func (s *Store) AddHostScans(scans []explorer.HostScan) error {
+func (s *Store) AddHostScans(backoffInterval time.Duration, scans []explorer.HostScan) error {
 	return s.transaction(func(tx *txn) error {
-		return addHostScans(tx, scans)
+		return addHostScans(tx, backoffInterval, scans)
```
n8maninger, line 1182: This looks like its probably fine to inline

persist/sqlite/consensus.go
```diff
@@ -1147,14 +1147,14 @@ func addHosts(tx *txn, hosts []explorer.Host) error {
 	return nil
 }
 
-func addHostScans(tx *txn, scans []explorer.HostScan) error {
-	unsuccessfulStmt, err := tx.Prepare(`UPDATE host_info SET last_scan = ?, last_scan_successful = 0, total_scans = total_scans + 1, failed_interactions = failed_interactions + 1 WHERE public_key = ?`)
+func addHostScans(tx *txn, backoffInterval time.Duration, scans []explorer.HostScan) error {
```
n8maninger, line 1150: More of a nit, but `NextScan time.Time` could be added to the `HostScan` instead of calculating it in the database. Saves a parameter and cleans up the logic here.

n8maninger (changes requested):

persist/sqlite/consensus.go
chris124567: If we wanted to do the calculation outside the query we'd need to retrieve the number of previous failures (`failed_interactions_streak`) for that host from the DB first which I feel would complicate things too much to be worth it

n8maninger: Does it complicate things because you feel the implementation is complicated or because it's outside the scope of this PR? How would you go about implementing this?

chris124567: Actually, we could retrieve the `failed_interactions_streak` count alongside the other details about a host in HostsForScanning.  Then it'd be pretty easy.

persist/sqlite/hosts.go
```diff
@@ -13,11 +13,11 @@ import (
 // HostsForScanning returns hosts ordered by their time to next scan.  Hosts
 // which are repeatedly offline will face an exponentially growing next scan
 // time to avoid wasting resources.
-// Note that only the PublicKey, V2, NetAddress, and V2NetAddresses fields are
-// populated.
+// Note that only the PublicKey, V2, NetAddress, V2NetAddresses,
```
n8maninger, line 16: If only those fields are populated, I’d consider a new struct.

n8maninger (approved):

PR #161, opened 2025-01-17
Rationalize search API HTTP error codes and add hosts and v2 contracts to search

chris124567:
Allow the search endpoint to see if a v2 contract with the specified ID or a host with the specified pubkey exists.  Ed25519 pubkeys and Hash256 values are both 32 bytes and we already strip prefixes on /search/:id so this works out.

Also, we now use errors that make more sense on the search endpoint.  Previously we would just 500 on all errors.  Now:
- If we get a malformed ID, return 400
- If we get an ID that we cannot find in the DB, return 404
- If we get an error in searching the DB, return 500
- Else return 200 and the search result

### DIFF
--- a/api/server.go
+++ b/api/server.go
@@ -729,6 +729,9 @@ func (s *server) searchIDHandler(jc jape.Context) {
 	if errors.Is(err, explorer.ErrNoSearchResults) {
 		jc.Error(ErrNoSearchResults, http.StatusNotFound)
 		return
+	} else if errors.Is(err, explorer.ErrSearchParse) {
+		jc.Error(err, http.StatusBadRequest)
+		return
 	} else if jc.Check("failed to search ID", err) != nil {
 		return
 	}
--- a/api/server.go
+++ b/api/server.go
@@ -2,7 +2,6 @@ package api
 
 import (
 	"context"
-	"encoding/hex"
 	"errors"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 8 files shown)


explorer/explorer.go
```diff
@@ -445,5 +449,19 @@ func (e *Explorer) Search(id types.Hash256) (SearchType, error) {
 		return SearchTypeContract, nil
 	}
 
-	return SearchTypeInvalid, errors.New("no such element")
+	v2Contracts, err := e.V2Contracts([]types.FileContractID{types.FileContractID(id)})
```
n8maninger, line 452: Good opportunity to clean this up by implementing it directly in the database. Will reduce IO since you're not loading data you're going to throw away and probably improve performance.

api/server.go
```diff
@@ -719,22 +719,26 @@ func (s *server) hostsHandler(jc jape.Context) {
...
+		jc.Error(errors.New("failed to decode hex"), http.StatusBadRequest)
+		return
+	} else if len(id) < idLen {
+		jc.Error(fmt.Errorf("expected length %d, got %d", idLen, len(id)), http.StatusBadRequest)
 		return
 	}
 
-	trunc := id[:maxLen]
+	trunc := id[:idLen]
 	result, err := s.e.Search(types.Hash256(trunc))
```
n8maninger, line 736: I'd probably make the search term a string instead since you have to support prefixed keys too.

n8maninger (changes requested):

peterjan (approved):

api/server.go
ChrisSchinnerl: Also please add a test to make sure we actually support it

n8maninger:
```go
// SearchTypeSiacoinElement means we found a contract with the given ID.
SearchTypeSiacoinElement SearchType = "siacoinElement"

// SearchTypeSiafundElement means we found a contract with the given ID.
SearchTypeSiafundElement SearchType = "siafundElement"
```

Copy pasta

ChrisSchinnerl (approved):

n8maninger (approved):

PR #176, opened 2025-02-28
Include time of next scan in hosts response

chris124567:
May be useful to users whose hosts have been down for a while to see when they will be scanned again due to the exponential backoff on failed scans, and it doesn't cost us anything to return this data.


n8maninger:
Is this ready for review?

PR #178, opened 2025-03-04
Add RHP4 fields to host metrics

chris124567:
/metrics/host only includes data from the RHP2 settings and RHP3 price table.  This adds averages for v2 hosts to the host metrics endpoint.


peterjan (approved):
Looks good to me

explorer/types.go
```diff
@@ -398,8 +398,9 @@ type HostMetrics struct {
 	// equal to TotalStorage - RemainingStorage)
 	RemainingStorage uint64 `json:"remainingStorage"`
 
-	Settings   rhpv2.HostSettings   `json:"settings"`
-	PriceTable rhpv3.HostPriceTable `json:"priceTable"`
+	Settings      rhpv2.HostSettings   `json:"settings"`
+	PriceTable    rhpv3.HostPriceTable `json:"priceTable"`
+	RHPV4Settings rhpv4.HostSettings   `json:"hostSettings"`
```
peterjan, line 403: Do we want to `omitempty` on `settings` and `hostSettings`? I also considered started a discussion on the naming but this is probably fine, `renterd` uses `settings` and `v2settings`.

ChrisSchinnerl: Unless this is already exposed anywhere else as `hostSettings` I'd probably indeed go for `v2Settings`. Seems a bit more straightforward to me.

Also should be checked for in some test to make sure it is set.

ChrisSchinnerl (changes requested):

ChrisSchinnerl (dismissed):

peterjan (approved):

PR #179, opened 2025-03-06
Add endpoint to search hosts by net address

chris124567:


explorer/explorer.go
```diff
@@ -402,6 +402,11 @@ func (e *Explorer) Hosts(pks []types.PublicKey) ([]Host, error) {
 	return e.s.QueryHosts(HostQuery{PublicKeys: pks}, HostSortPublicKey, HostSortAsc, 0, math.MaxInt64)
 }
 
+// HostNetAddresses returns the hosts with the specified net addressess.
+func (e *Explorer) HostNetAddresses(netAddresses []string) ([]Host, error) {
+	return e.s.QueryHosts(HostQuery{NetAddresses: netAddresses}, HostSortPublicKey, HostSortAsc, 0, math.MaxInt64)
```
peterjan, line 407: I think this method could have a little more logic to deal with no hosts being found, or potentially handle the case where more than one host is found. E.g. we could limit by one and perform the `len(hosts)` check here.

persist/sqlite/hosts.go
```diff
@@ -65,115 +65,152 @@ func (s *Store) HostsForScanning(minLastAnnouncement time.Time, limit uint64) (r
...
+		case explorer.HostSortDownloadPrice:
+			sortColumn = "CASE WHEN v2=1 THEN rhp4_prices_egress_price ELSE settings_download_bandwidth_price END"
+		case explorer.HostSortUploadPrice:
+			sortColumn = "CASE WHEN v2=1 THEN rhp4_prices_ingress_price ELSE settings_upload_bandwidth_price END"
+		case explorer.HostSortUsedStorage:
+			sortColumn = "CASE WHEN v2=1 THEN rhp4_settings_used_storage ELSE settings_used_storage END"
+		case explorer.HostSortTotalStorage:
+			sortColumn = "CASE WHEN v2=1 THEN rhp4_settings_total_storage ELSE settings_total_storage END"
+		default:
+			return fmt.Errorf("invalid sort column: %s", sortBy)
```
peterjan, line 199: This should probably be an exported error we can check for to properly return a 400 status code when it happens.

api/server.go
```diff
@@ -719,6 +720,21 @@ func (s *server) hostsHandler(jc jape.Context) {
...
+		return
+	}
+	hosts, err := s.e.HostNetAddresses([]string{netAddress})
+	if jc.Check("failed to get host", err) != nil {
+		return
+	} else if len(hosts) == 0 {
+		jc.Error(ErrHostNotFound, http.StatusNotFound)
+		return
+	}
+	jc.Encode(hosts[0])
```
ChrisSchinnerl, line 735: Do we need a new endpoint? We could extend the one that queries hosts with this functionality.

It's also not safe to assume that this always returns a single host. If shut down my host, create a new one on the same server and then announce the same address, two hosts will have the same address in the database.

ChrisSchinnerl (changes requested):

api/server.go
chris124567: Yeah, I guess we don't need a special endpoint.  Unlike pubkeys, net addresses aren't so unique.  With /hosts clients can just submit `{"netAddresses": ["x.x.x.x:1234", ...]}`.

api/server.go
```diff
@@ -74,6 +74,7 @@ type (
 		Search(id string) (explorer.SearchType, error)
 
 		Hosts(pks []types.PublicKey) ([]explorer.Host, error)
+		HostNetAddresses(netAddresses []string) ([]explorer.Host, error)
```
ChrisSchinnerl, line 77: This isn't needed anymore right?

ChrisSchinnerl (changes requested):
Can we also get an integration test that fetches hosts by net address via the http API?

ChrisSchinnerl (approved):

peterjan (approved):

PR #181, opened 2025-03-09
Add v2 numbers to total and remaining storage in metrics

chris124567:


persist/sqlite/metrics.go
```diff
@@ -66,7 +66,9 @@ func (s *Store) HostMetrics() (result explorer.HostMetrics, err error) {
 			}
 
 			result.TotalStorage += host.Settings.TotalStorage
+			result.TotalStorage += host.V2Settings.TotalStorage
 			result.RemainingStorage += host.Settings.RemainingStorage
+			result.RemainingStorage += host.V2Settings.RemainingStorage
```
ChrisSchinnerl, line 71: Make sure to update any tests to check those additional fields and also API docs.
Especially since `RemainingStorage` uses different units for v1 and v2. The latter uses sectors afaik. Would also double check `TotalStorage`.

ChrisSchinnerl (changes requested):

persist/sqlite/metrics.go
```diff
@@ -66,7 +68,9 @@ func (s *Store) HostMetrics() (result explorer.HostMetrics, err error) {
 			}
 
 			result.TotalStorage += host.Settings.TotalStorage
+			result.TotalStorage += proto4.SectorSize * host.V2Settings.TotalStorage
```
peterjan, line 71: `TotalStorage` and `RemainingStorage` should already expressed in bytes. Shouldn't these totals be moved down inside into the `if !host.V2 {` blocks? It seems like the remaining and total storage of a V2 hosts would count twice towards the total if it has both `Settings` populated.

persist/sqlite/metrics.go
```diff
@@ -66,7 +68,9 @@ func (s *Store) HostMetrics() (result explorer.HostMetrics, err error) {
...
 			result.RemainingStorage += host.Settings.RemainingStorage
+			result.RemainingStorage += proto4.SectorSize * host.V2Settings.RemainingStorage
 
 			if !host.V2 {
```
peterjan, line 75: A small nitpick, but it might be clearer to flip this and frame it with hosts being on V2 as the default. I wouldn’t normally mention it, but it’d be nice if all our new code approached things from that perspective first and foremost.

persist/sqlite/metrics.go
chris124567: We sort of do this here https://github.com/SiaFoundation/explored/pull/181/commits/542f63bafa2f80850018c0a9e5ae8acae9be1659#diff-f1c64d840d659e37e8885562cee69016a811979ab014f064ea6488baea07d263R362

peterjan: It appears `sia1.euregiohosting.nl` is offline at the moment, causing `TestScan` to fail for me locally. What's the reason we include an actual live host in these tests? We also appear to be sleeping for quite long periods of time, e.g. `time.Sleep(4 * cfg.Timeout)` that's a two minute sleep

ChrisSchinnerl: @chris124567 can you rewrite that test to use ephemeral hosts you spin up yourself rather than live hosts? 
Try to keep it as minimal as possible. e.g. you shouldn't need a full host but only a mocked one which can give you an announcement for itself and listen for RPCSettings and RPCPriceTable requests. Then you can freely pick the settings it should return and check for them rather than relying on a live host.

ChrisSchinnerl (approved):

persist/sqlite/scan_test.go
```diff
@@ -221,88 +388,90 @@ func TestScan(t *testing.T) {
 		}
 	}
 
-	{
-		now := types.CurrentTimestamp()
-		lastAnnouncementCutoff := now.Add(-cfg.MinLastAnnouncement)
+	time.Sleep(1 * cfg.Timeout)
```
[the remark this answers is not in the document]
chris124567: We could probably shorten the duration a little bit but the explorer has a thread that calls HostsForScanning in a loop (it sleeps for 15 seconds if there were no results).  If HostsForScanning returns hosts to be scanned, then each host gets a goroutine to scan it which is not instantaneous.  We could loop in the test until HostsForScanning returns 0, but if there's a bug then that could go on forever so we would probably also want to have some sort of timeout too.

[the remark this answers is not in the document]
ChrisSchinnerl: I'd configure it to be much shorter for tests. 15s is still pretty rough for something that could probably happen every 100ms or so in testing.

persist/sqlite/scan_test.go
```diff
@@ -121,52 +232,108 @@ func TestScan(t *testing.T) {
...
+	pubkey4 := pk4.PublicKey()
+
+	rhp2Addr, _ := testV1Host(t, pk1, &settings, &table)
+	v4Addr, _ := testV2Host(t, pk3, cm, s, w, c, sr, ss, zap.NewNop())
 
 	cfg := config.Scanner{
-		Threads:             10,
-		Timeout:             30 * time.Second,
+		BatchSize:           100,
+		Timeout:             10 * time.Second,
```
peterjan, line 335: This is still a very high number. The first thing `TestScan` does is wait 10s to sync up. Likely because of

```
func (e *Explorer) waitForSync() error {
	ticker := time.NewTicker(10 * time.Second)
	defer ticker.Stop()
```

`nit` in the logging I noticed `logger.go:146: 2025-03-19T09:14:05.779+0100 INFO    shutdown:       {"error": "context canceled"}`, which is not really a useful log line. If anything it could/should be at the `debug` level. In `renterd` we weren't great about this either and we get frequent complaints about the size of our `renterd.log`. In approach in indexd is to info log only when the error is actionable.

config/config.go
```diff
@@ -18,9 +18,24 @@ type (
 
 	// Scanner contains the configuration for the host scanner.
 	Scanner struct {
-		Threads             int           `yaml:"threads,omitempty"`
-		Timeout             time.Duration `yaml:"timeout,omitempty"`
-		MaxLastScan         time.Duration `yaml:"maxLastScan,omitempty"`
+		// BatchSize represents the maximum number of hosts we will
```
peterjan, line 21: The docstrings are great, I like the field names less. 

`BatchSize` -> `NumThreads`: since every unscanned host gets its own goroutine, threads cover the load better
`Timeout` -> `ScanTimeout`: doesn't really matter
`CheckAgainDelay` -> `ScanFrequency`: doesn't really matter
`MaxLastScan` -> `ScanInterval`: I didn't really get the `Max` part? There's no `LastScan` field and it's not even the max possible value because we have an exp backoff on failures. It's different from `MinLastAnnouncement` which clearly gets used as a cutoff for the `LastAnnouncement` column.

Another one I noticed is `FailedInteractionsStreak`, which we would usually call `ConsecutiveFailedScans`, which is also how you describe it in the docstring of `MaxLastScan`.

ChrisSchinnerl (approved):

cmd/explored/main.go
```diff
@@ -45,8 +45,9 @@ var cfg = config.Config{
 		EnableUPNP: false,
 	},
 	Scanner: config.Scanner{
-		Threads:             10,
+		BatchSize:           100,
```
peterjan, line 48: Going to leave this here but in `NewExplorer(cm ChainManager, store Store, batchSize int, scanCfg config.Scanner, log *zap.Logger) ` you could pass the indexerCfg instead of only the batch size to make it more consistent. This is partly why I suggest to rename `batchSize` because there's essentially two batchsizes and we can't immediately tell which one's used where.

peterjan:
@chris124567 feel free to resolve my comments and/or tackle some of the remarks in a F/U, I do feel there's no point in injecting a 100ms scan frequency if the test sleeps for 10 second on a chain sync. In general I'm just trying to ensure we are somewhat consistent between our repos.

peterjan (approved):

cmd/explored/main.go
ChrisSchinnerl, line 1: I'd address PJ's remarks in one or more follow up PRs. Having consistent naming between nodes is definitely nice to have.

I'd also address the scanning test being slow in a separate PR to split concerns a bit. Most tests pass in less than 1s but that one takes forever even though there is really no need for that.

persist/sqlite/scan_test.go
chris124567: Resolved in followup

cmd/explored/main.go
chris124567: Resolved in followup

PR #184, opened 2025-03-11
Prevent scanning functions from hanging on some v1 hosts

chris124567:
If a host responds but with an incompatible protocol (or just extremely slowly), the RHP code will wait for a response of a certain size indefinitely.  This PR adds connection deadlines to prevent that from happening.  This fixes the low host count on beta.siascan.com that has happened because some scanning goroutines never finished, so the WaitGroup never finished either, which meant scans stopped being added to the database.

A similar issue with v2 hosts came up lately that was fixed by upgrading coreutils: https://github.com/SiaFoundation/explored/pull/180

### DIFF
--- a/explorer/scan.go
+++ b/explorer/scan.go
@@ -54,18 +54,12 @@ func (e *Explorer) waitForSync() error {
 	return nil
 }
 
-func rhpv2Settings(ctx context.Context, publicKey types.PublicKey, netAddress string) (crhpv2.HostSettings, error) {
+func rhpv2Settings(ctx context.Context, deadline time.Time, publicKey types.PublicKey, netAddress string) (crhpv2.HostSettings, error) {
 	conn, err := (&net.Dialer{}).DialContext(ctx, "tcp", netAddress)
 	if err != nil {
 		return crhpv2.HostSettings{}, fmt.Errorf("failed to connect to host: %w", err)
 	}
 	defer conn.Close()
-
-	// default timeout if context doesn't have one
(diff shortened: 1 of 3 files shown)


explorer/scan.go
```diff
@@ -54,18 +54,12 @@ func (e *Explorer) waitForSync() error {
 	return nil
 }
 
-func rhpv2Settings(ctx context.Context, publicKey types.PublicKey, netAddress string) (crhpv2.HostSettings, error) {
+func rhpv2Settings(ctx context.Context, deadline time.Time, publicKey types.PublicKey, netAddress string) (crhpv2.HostSettings, error) {
```
n8maninger, line 57: instead of adding a new parameter, set a default of like 30 seconds and override with ctx.Deadline() if set.

explorer/scan.go
```diff
@@ -83,18 +77,12 @@ func rhpv2Settings(ctx context.Context, publicKey types.PublicKey, netAddress st
 	return settings, nil
 }
 
-func rhpv3PriceTable(ctx context.Context, publicKey types.PublicKey, netAddress string) (priceTable crhpv3.HostPriceTable, err error) {
+func rhpv3PriceTable(ctx context.Context, deadline time.Time, publicKey types.PublicKey, netAddress string) (priceTable crhpv3.HostPriceTable, err error) {
```
n8maninger, line 80: ditto

n8maninger (approved):

Issue #186, opened 2025-03-15
Add host scan endpoint


CtrlAltDefeat94:
It would be useful to trigger a host scan on demand. 
Renterd already has this endpoint; https://api.sia.tech/renterd#3faf99ac-a7f0-49b6-9e51-e2c03421dd89

chris124567:
We can set the next scan time for a host to the current timestamp to basically ensure that the host will be scanned in the very near future.  One thing that sort of worries me is someone could use this in a really annoying way and overwhelm the scanning queue.  Hosts can already tell roughly when they will be scanned next with the `NextScan` field in the response.  I suppose we can do this but it might end up being disabled on the public instance.

ChrisSchinnerl:
It can certainly be useful to have an endpoint that performs a blocking scan on a host immediately. e.g. when you run your own host, it appears offline, you forward some ports and want to confirm that everything is working now.
However, it's not something you'd want to give the general public unrestricted access to. 

That said it's probably fine to add it for people who run their own explorer for now and later think about potentially exposing it with a per-host rate limit.

CtrlAltDefeat94:
I hadnt thought about the possible abuse cases.

Not sure if it's difficult to implement, but a config option to only allow the endpoint from localhost could help.

PR #189, opened 2025-03-18
Add host scan endpoint

chris124567:
https://github.com/siafoundation/explored/issues/186

- Add endpoint to scan host protected by basic auth (set password with environment variable `MANUAL_SCAN_PASSWORD`)
- Add field LastScanError to host which represents the error we got scanning the host if there was one
- Refactor and simplify scan test + add check to make sure manual scan works


CtrlAltDefeat94:
Since it's (somewhat) related I'm asking here as it would save a PR. Could the reason behind a failed scan also be added to the API output?
I can make another issue/FR if you want to keep things separately or add it later :)

chris124567:
We can add it.  Right now if there's an error it would be present in the response but not formatted, because I sort of envisioned this endpoint being manually `curl`ed as needed.

n8maninger:
Definitely want a manual scan behind basic auth

CtrlAltDefeat94:
> We can add that. Right now if there's an error it would be present in the response but not formatted with JSON or anything, because I sort of envisioned this endpoint being manually `curl`ed as needed.

I would also want to know why the most recent automated scan failed. Currently I think it's only possible to figure why from the (debug) logging.

api/client.go
```diff
@@ -255,6 +255,12 @@ func (c *Client) Host(key types.PublicKey) (resp explorer.Host, err error) {
 	return
 }
 
+// TriggerHostScan triggers a manual host scan.
+func (c *Client) TriggerHostScan(key types.PublicKey) (err error) {
+	err = c.c.POST(fmt.Sprintf("/pubkey/%s/host/scan", key), nil, nil)
```
peterjan, line 260: I would've expected `POST /host/:hostkey/scan`. We have `POST /hosts` and we also have `/contracts/:id`, what's the benefit of `/pubkey/:key` over `/host/:hostkey`?


```
		"GET    /contracts/:id/revisions": srv.contractsIDRevisionsHandler,
		"POST   /contracts":               srv.contractsBatchHandler,

		"GET    /v2/contracts/:id":           srv.v2ContractsIDHandler,
		"GET    /v2/contracts/:id/revisions": srv.v2ContractsIDRevisionsHandler,
		"POST   /v2/contracts":               srv.v2ContractsBatchHandler,

		"GET    /v2/pubkey/:key/contracts": srv.v2PubkeyContractsHandler,

		"GET    /pubkey/:key/contracts": srv.pubkeyContractsHandler,
		"GET    /pubkey/:key/host":      srv.pubkeyHostHandler,
		"POST   /pubkey/:key/host/scan": srv.pubkeyHostScanHandler,

		"GET    /metrics/block":     srv.blocksMetricsHandler,
		"GET    /metrics/block/:id": srv.blocksMetricsIDHandler,
		"GET    /metrics/host":      srv.hostMetricsHandler,

		"POST   /hosts": srv.hostsHandler,
```

api/server.go
```diff
@@ -73,6 +73,7 @@ type (
 		V2ContractRevisions(id types.FileContractID) (result []explorer.V2FileContract, err error)
 		Search(id string) (explorer.SearchType, error)
 
+		TriggerHostScan(pk types.PublicKey) error
```
peterjan, line 76: `nit` triggering (for me) implies a potentially asynchronous action, would rename it to `ScanHost`

cmd/explored/main.go
```diff
@@ -61,7 +61,7 @@ var cfg = config.Config{
 		BatchSize: 1000,
 	},
 	Log: config.Log{
-		Level: "info",
+		Level: "debug",
```
peterjan, line 64: is this a debug leftover or do we really want to change the default log level to `debug`, we normally have it at `info`

cmd/explored/main.go
```diff
@@ -298,7 +298,8 @@ func runRootCmd(ctx context.Context, log *zap.Logger) error {
 	}
 	go ex.Start(ctx)
 
-	api := api.NewServer(e, cm, s, ex)
+	manualScanPassword := os.Getenv("MANUAL_SCAN_PASSWORD")
```
peterjan, line 301: we already have `EXPLORED_CONFIG_FILE`, so should we use `EXPLORED_HOST_SCAN_PASSWORD` or something that mentions `AUTH_` perhaps?

api/server.go
```diff
@@ -691,6 +693,25 @@ func (s *server) pubkeyHostHandler(jc jape.Context) {
 	jc.Encode(hosts[0])
 }
 
+func (s *server) pubkeyHostScanHandler(jc jape.Context) {
+	// We could use jape.BasicAuth when defining the route in the map, but it
+	// makes the jape linter think that the route is undefined, so we have some
+	// auth code here.
+	if _, p, ok := jc.Request.BasicAuth(); !ok || s.manualScanPassword == "" || p != s.manualScanPassword {
```
peterjan, line 700: In `walletd` we define `wrapAuthHandler`, would that be a solution to the linting problem? Not sure I fully understand why the jape linter would think the route is undefined, never encountered that before myself.

persist/sqlite/scan_test.go
```diff
@@ -366,121 +353,60 @@ func TestScan(t *testing.T) {
...
-	for {
-		tip, err := e.Tip()
-		if err != nil {
-			t.Fatal(err)
-		}
-		if tip != cm.Tip() {
-			time.Sleep(time.Second)
-		} else {
-			break
+	waitForTip := func() {
```
peterjan, line 377: mark it as a helper using `t.Helper()`

api/server.go
```diff
@@ -691,6 +693,25 @@ func (s *server) pubkeyHostHandler(jc jape.Context) {
...
+		jc.Error(errors.New("auth needed for manual scan"), http.StatusUnauthorized)
+		return
+	}
+
+	var key types.PublicKey
+	if jc.DecodeParam("key", &key) != nil {
+		return
+	}
+
+	if jc.Check("failed to scan host", s.e.TriggerHostScan(key)) != nil {
```
peterjan, line 710: does it make sense to return the `HostScan` if it succeeded?

explorer/explorer.go
```diff
@@ -412,6 +421,52 @@ func (e *Explorer) QueryHosts(params HostQuery, sortBy HostSortColumn, dir HostS
 	return e.s.QueryHosts(params, sortBy, dir, offset, limit)
 }
 
+// TriggerHostScan triggers a manual host scan.
+func (e *Explorer) TriggerHostScan(pk types.PublicKey) error {
+	hosts, err := e.Hosts([]types.PublicKey{pk})
```
peterjan, line 426: could use variadic parameter

explorer/explorer.go
```diff
@@ -412,6 +421,52 @@ func (e *Explorer) QueryHosts(params HostQuery, sortBy HostSortColumn, dir HostS
...
+		}
+	} else {
+		e.log.Debug("manual host scan succeeded", zap.Stringer("pk", host.PublicKey))
+	}
+	// We don't apply the exponential delay penalty to manually scanned hosts.
+	// Given that this would mostly be used by someone setting up or
+	// configuring their host, it seems wrong to use it here.
+	scan.NextScan = now.Add(e.scanCfg.ScanInterval)
+
+	if err := e.s.AddHostScans([]HostScan{scan}); err != nil {
```
peterjan, line 464: could use variadic parameter

cmd/explored/main.go
n8maninger: I’d stay consistent with the other apps and go `EXPLORED_API_PASSWORD`

api/server.go
```diff
@@ -760,14 +781,16 @@ func (s *server) exchangeRateHandler(jc jape.Context) {
...
-		cm:        cm,
-		e:         e,
-		s:         s,
-		ex:        ex,
-		startTime: time.Now().UTC(),
+		cm:                 cm,
+		e:                  e,
+		s:                  s,
+		ex:                 ex,
+		manualScanPassword: manualScanPassword,
```
ChrisSchinnerl, line 790: I'd just call this `apiPassword`. We probably also want to protect the endpoint for connecting to peers now that I'm looking at the routes. Because letting someone else decide what peers you are connected to isn't very secure.

ChrisSchinnerl (changes requested):

cmd/explored/main.go
```diff
@@ -298,7 +298,8 @@ func runRootCmd(ctx context.Context, log *zap.Logger) error {
 	}
 	go ex.Start(ctx)
 
-	api := api.NewServer(e, cm, s, ex)
+	apiPassword := os.Getenv("EXPLORED_API_PASSWORD")
```
n8maninger, line 301: Add this to the config
```go
var cfg = config.Config{
	Directory: ".",
	HTTP: config.HTTP{
		Address: ":9980",
        Password: os.Getenv("EXPLORED_API_PASSWORD")
	},
}
```

explorer/types.go
```diff
@@ -330,6 +330,7 @@ type HostScan struct {
 	PublicKey types.PublicKey `json:"publicKey"`
 	Location  geoip.Location  `json:"location"`
 	Success   bool            `json:"success"`
+	Error     string          `json:"error"`
```
n8maninger, line 333: Probably prefer `error` so that you can check for `nil` rather than `""`.

explorer/types.go
```diff
@@ -377,6 +378,7 @@ type Host struct {
 	KnownSince             time.Time `json:"knownSince"`
 	LastScan               time.Time `json:"lastScan"`
 	LastScanSuccessful     bool      `json:"lastScanSuccessful"`
+	LastScanError          string    `json:"lastScanError"`
```
n8maninger, line 381: Prefer `*string` or `error` so that it's `nil` rather than `""`. The problem with `error` is that you would need a wrapping type for JSON.

chris124567: I went with *string in both.  We return the HostScan as the response for /hosts/%s/scan so we can't use `error` there.

api/api_test.go
```diff
@@ -79,7 +79,7 @@ func newServer(t *testing.T, cm *chain.Manager, e *explorer.Explorer, listenAddr
 		exchangerates.CurrencyUSD: exchangerates.KrakenPairSiacoinUSD}, time.Second)
 	go ex.Start(ctx)
 
-	api := api.NewServer(e, cm, &syncer.Syncer{}, ex)
+	api := api.NewServer(e, cm, &syncer.Syncer{}, ex, "")
```
ChrisSchinnerl, line 82: I'd set a hardcoded test password here rather than no password.

api/api_test.go
ChrisSchinnerl, line 1: We should at least have a test that hits the authenticated endpoints without an authenticated client and makes sure that fails with the right error.

ChrisSchinnerl (dismissed):
Can be a f/u but the other endpoint still needs to have a test to verify it's authenticated

api/api_test.go
```diff
@@ -608,12 +609,17 @@ func TestAPI(t *testing.T) {
...
 			pubkey := pk1.PublicKey()
-			badAuthClient := api.NewClient("http://"+listenAddr+"/api", "")
 			_, err := badAuthClient.ScanHost(pubkey)
 			if err == nil {
 				t.Fatal("got no error when trying to use manual scan with bad auth")
 			}
 		}},
+		{"Syncer connect with invalid credential", func(t *testing.T) {
+			err := badAuthClient.SyncerConnect("127.0.0.0.1:65535")
+			if err == nil {
```
ChrisSchinnerl, line 619: When making sure that something fails for a given reason, make sure to check what the reason was. Either by checking against a specific exposed error, status code (should be 401) or at least a string comparison if the error is not exposed.

api/server.go
```diff
@@ -114,14 +115,26 @@ var (
...
 
 	startTime time.Time
 }
 
+func (s *server) checkAuth(jc jape.Context) bool {
+	// We could use jape.BasicAuth when defining the route in the map, but it
+	// makes the jape linter think that the route is undefined, so we have some
+	// auth code here.
+	if _, p, ok := jc.Request.BasicAuth(); !ok || s.apiPassword == "" || p != s.apiPassword {
+		jc.Error(errors.New("auth needed for manual scan"), http.StatusUnauthorized)
```
ChrisSchinnerl, line 132: This error message is no longer accurate. Could be something more generic.

ChrisSchinnerl (changes requested):

api/server.go
peterjan: Ok. Took a quick look and I agree it is annoying. I tried splitting it in authenticated and unauthenticated routes to then merge the two routers, which works but doesn't solve the problem because jape loses track of all handlers 😁

```
authRoutes := auth(apiPassword)(jape.Mux(map[string]jape.Handler{
		"POST   /hosts/:key/scan":        srv.pubkeyHostScanHandler,
		"POST   /syncer/broadcast/block": srv.syncerBroadcastBlockHandler,
	}))
```

api/server.go
```diff
@@ -89,6 +89,10 @@ const (
 )
 
 var (
+	// ErrInvalidAuth is returned when the supplied credentials for a protected
+	// endpoint are invalid.
+	ErrInvalidAuth = errors.New("invalid auth")
```
peterjan, line 94: `nit` this error is user facing and one they're very likely to see, so "invalid auth" might be a bit crude, maybe one of the following "bad credentials", "authentication failed", "unauthorized: bad credentials"?

peterjan (approved):

PR #190, opened 2025-03-20
Fix V2 storage total followups

chris124567:
- Rename scanner config fields
- Make scanning goroutine shutdown log use debug mode instead of info mode
- Reduce the 10 second interval in waitForSync to 1 second
- Lower wait in test to 3 seconds from 10 seconds

Follow up to https://github.com/SiaFoundation/explored/pull/181


ChrisSchinnerl (approved):
The sleep is better now but I wonder if we could further reduce it to <=1s. The test still takes 7 seconds to run right now.

explorer/scan.go
```diff
@@ -23,7 +23,7 @@ func isSynced(b Block) bool {
 }
 
 func (e *Explorer) waitForSync() error {
-	ticker := time.NewTicker(10 * time.Second)
+	ticker := time.NewTicker(1 * time.Second)
```
peterjan, line 26: This is fine if we really want to check every second. We don't generally tweak production parameters to make tests run faster.

chris124567: Reindexing from scratch takes 1-2 days, so it is not necessary to check this frequently.  But even with SQLite being single threaded I doubt it costs very much to lower it to 1s.

chris124567:
Ok, latest commit gets it down to 4 seconds.

chris124567:
Alright I think low single digit seconds is a reasonable amount of time.  I think to get to < 1 second would require a nontrivial amount of changes and that probably should not a priority right now.

peterjan (approved):

ChrisSchinnerl (approved):

PR #192, opened 2025-03-27
Expose type of resolution for v2 contracts

chris124567:
Add field to `V2FileContract` which represents the type of resolution, if there was one.  It can either be a `V2ResolutionRenewal`, `V2ResolutionStorageProof`, or `V2ResolutionExpiration`.  This saves consumers from having to retrieve the transaction using the resolution transaction ID to figure out the resolution type.

### DIFF
--- a/explorer/events.go
+++ b/explorer/events.go
@@ -353,7 +353,6 @@ func CoreToExplorerV2Transaction(txn types.V2Transaction) (result V2Transaction)
 		parent.V2FileContractElement.StateElement = fcr.Parent.StateElement
 
 		var res any
-		var typ ResolutionType
 		switch v := fcr.Resolution.(type) {
 		case *types.V2FileContractRenewal:
 			res = V2FileContractRenewal{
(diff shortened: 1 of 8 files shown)


ChrisSchinnerl (approved):

chris124567:
@telestrial PS with this line

https://github.com/SiaFoundation/explored/pull/192/files#diff-f94362c60d1db34c0f564cf8877be856fb753444969de65ba7124fcaebdc04f7L254

The JSON tag for the Type field on V2FileContractResolution was renamed to "type" from "string" since that it what it actually is.

https://github.com/SiaFoundation/explored/pull/192/files#diff-f94362c60d1db34c0f564cf8877be856fb753444969de65ba7124fcaebdc04f7R125

Also, when we marshal the type it will be "storage_proof" instead of "storageProof" ("renewal" and "expiration" stay the same) to be more in line with the snake_case naming used when marshaling Sources to JSON (i.e. "miner_payout").

ChrisSchinnerl (approved):

peterjan (approved):

PR #194, opened 2025-04-01
Don't wait for explorer to catch up before starting

chris124567:
Resolve issue where exchange rates (or any API endpoint) would timeout when explorer was given an existing consensus DB because NewExplorer was blocking on indexing chain data to the current chain manager tip.  Manually tested

### DIFF
--- a/persist/sqlite/scan_test.go
+++ b/persist/sqlite/scan_test.go
@@ -377,20 +377,13 @@ func TestScan(t *testing.T) {
 	waitForTip := func() {
 		t.Helper()
 
-		const duration = time.Second
 		for {
-			tip, err := e.Tip()
-			if errors.Is(err, explorer.ErrNoTip) {
-				time.Sleep(duration)
-				continue
-			} else if err != nil {
+			if tip, err := e.Tip(); err != nil && !errors.Is(err, explorer.ErrNoTip) {
 				t.Fatal(err)
-			}
-			if tip != cm.Tip() {
-				time.Sleep(duration)
-			} else {
+			} else if tip == cm.Tip() {
 				break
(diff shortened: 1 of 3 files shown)


ChrisSchinnerl (approved):

peterjan:
👀 conflicts

persist/sqlite/scan_test.go
```diff
@@ -395,7 +395,9 @@ func TestScan(t *testing.T) {
 
 	for {
 		tip, err := e.Tip()
-		if err != nil {
+		if err == explorer.ErrNoTip {
```
peterjan, line 398: `nit` use errors.Is()

peterjan, line 398: does `ErrNoTip` make sense? we could just return `types.ChainIndex{}, nil`?

persist/sqlite/scan_test.go
```diff
@@ -376,13 +376,18 @@ func TestScan(t *testing.T) {
 
 	waitForTip := func() {
 		t.Helper()
+
+		const duration = time.Second
 		for {
 			tip, err := e.Tip()
-			if err != nil {
+			if errors.Is(err, explorer.ErrNoTip) {
```
peterjan, line 383:
```go
for {
    if tip, err := e.Tip(); err != nil && !errors.Is(err, explorer.ErrNoTip) {
        t.Fatal(err)
    } else if tip == cm.Tip() {
        break
    }
    time.Sleep(time.Second)
}
```

```

peterjan (approved):

ChrisSchinnerl (approved):

PR #198, opened 2025-04-07
Add renewed from / renewed to contract for v2 contracts

chris124567:
Add fields to `V2Contract` type representing the ID of the previous contract that renewed to the contract (`RenewedFromID`), and `RenewedToID`, the ID the new renewal contract.


explorer/types.go
```diff
@@ -268,6 +268,9 @@ type Transaction struct {
 type V2FileContract struct {
 	TransactionID types.TransactionID `json:"transactionID"`
 
+	RenewedFromID *types.FileContractID `json:"renewedFromID"`
```
peterjan, line 271: we generally use `RenewedTo` and `RenewedFrom`

persist/sqlite/v2consensus_test.go
```diff
@@ -1385,7 +1398,283 @@ func TestV2FileContractResolution(t *testing.T) {
...
+	v2FC0ID := txn1.V2FileContractID(txn1.ID(), 0)
+	{
+		fcs, err := db.V2Contracts([]types.FileContractID{v2FC0ID})
+		if err != nil {
+			t.Fatal(err)
+		}
+		testutil.CheckV2FC(t, txn1.FileContracts[0], fcs[0])
+	}
+
+	// Check that they it is not resolved and does not have RenewedTo set
```
peterjan, line 1483: `nit` typo in the comment

persist/sqlite/v2consensus_test.go
```diff
@@ -1385,7 +1398,283 @@ func TestV2FileContractResolution(t *testing.T) {
...
+		}},
+		FileContracts: []types.V2FileContract{v2FC},
+	}
+	testutil.SignV2TransactionWithContracts(cm.TipState(), pk1, renterPrivateKey, hostPrivateKey, &txn1)
+
+	b1 := testutil.MineV2Block(cm.TipState(), []types.V2Transaction{txn1}, types.VoidAddress)
+	if err := cm.AddBlocks([]types.Block{b1}); err != nil {
+		t.Fatal(err)
+	}
+	syncDB(t, db, cm)
```
peterjan, line 1470: It takes 50 lines to create a store and add a test contract, I'm not sure how much this whole block is repeated but I bet it's something we need quite often. Perhaps we should F/U with a PR that introduces a test util that allows easily creating and renewing contracts.

persist/sqlite/v2consensus.go
```diff
@@ -264,28 +264,44 @@ func updateV2FileContractElements(tx *txn, revert bool, index types.ChainIndex,
...
+	if err != nil {
+		return fmt.Errorf("updateV2FileContractIndices: failed to prepare renewed from statement: %w", err)
+	}
+	defer renewedFromStmt.Close()
+
 	for _, update := range fces {
 		// id stays the same even if revert happens so we don't need to check that here
 		fcID := update.FileContractElement.ID
 
 		if revert {
```
peterjan, line 283: I'm assuming it's fine because of the tests you added, but could you explain a bit why the revert branch doesn't look like a mirrored copy from the apply branch? You'd expect a revert to exactly revert the apply right? so how can it be done in one statement while apply needs two statements?

persist/sqlite/v2consensus.go
```diff
@@ -264,28 +264,44 @@ func updateV2FileContractElements(tx *txn, revert bool, index types.ChainIndex,
...
 
 func updateV2FileContractIndices(tx *txn, revert bool, index types.ChainIndex, fces []explorer.V2FileContractUpdate) error {
-	resolutionIndexStmt, err := tx.Prepare(`UPDATE v2_last_contract_revision SET resolution_type = ?, resolution_height = ?, resolution_block_id = ?, resolution_transaction_id = ? WHERE contract_id = ?`)
+	resolutionIndexStmt, err := tx.Prepare(`UPDATE v2_last_contract_revision SET resolution_type = ?, resolution_height = ?, resolution_block_id = ?, resolution_transaction_id = ?, renewed_to_id = ? WHERE contract_id = ?`)
 	if err != nil {
 		return fmt.Errorf("updateV2FileContractIndices: failed to prepare resolution index statement: %w", err)
 	}
 	defer resolutionIndexStmt.Close()
 
+	renewedFromStmt, err := tx.Prepare(`UPDATE v2_last_contract_revision SET renewed_from_id = ? WHERE contract_id = ?`)
```
peterjan, line 273: I noticed `contract_id` is sometimes an integer referencing an id of another table, but it can also be a `BLOB` holding the file contract id. Now we're adding `renewed_from_id` and `renewed_to_id` which for me make the scheme a little more confusing over all. I'd personally rename it `renewed_from` and `renewed_to` but I'm also fine with keeping it the way it is because the `_id` "inconsistencies" are already there.

peterjan (approved):

persist/sqlite/init.sql
ChrisSchinnerl, line 1: The PR looks good but it is breaking the database schema and requires a resync.
Since we are running this on mainnet now we can't just break production by merging this.

We got two options here.
1. migration code
2. syncing a "staging" explored which we then swap in for the actual one

I got a call with Nate later this afternoon so I can see if he has a strong opinion on this before we hit v1 but the latter approach might work fine until we consider the explorer stable.

peterjan (approved):

ChrisSchinnerl (approved):

PR #200, opened 2025-04-07
Set merkle proofs when retrieving contract/output elements

chris124567:


api/server.go
```diff
@@ -870,6 +884,8 @@ func NewServer(e Explorer, cm ChainManager, s Syncer, ex exchangerates.Source, a
 
 		"POST   /hosts": srv.hostsHandler,
 
+		"GET    /proof/:index": srv.proofIndexHandler,
```
peterjan, line 887: This route is not terribly descriptive. It definitely short and sweet but it's not very user friendly. Should we consider `leaves/:index/proof` or `/state/tree/proofs/:index`? Feel free to resolve though if you disagree.

peterjan (approved):

api/server.go
```diff
@@ -763,6 +764,19 @@ func (s *server) hostsHandler(jc jape.Context) {
 	jc.Encode(hosts)
 }
 
+func (s *server) proofIndexHandler(jc jape.Context) {
```
ChrisSchinnerl, line 767: Do we need this? Don't we have an endpoint that already lets you fetch a file contract including its state element? It would be a bit more usable to have an endpoint that you can fetch a contract from with its up-to-date proof instead of having to call another endpoint.

chris124567: I've been leaving them as nil because retrieving them requires multiple queries.  I guess at least for these endpoints:
```
		"GET    /outputs/siacoin/:id": srv.outputsSiacoinHandler,
		"GET    /outputs/siafund/:id": srv.outputsSiafundHandler,
		"GET    /v2/contracts/:id":           srv.v2ContractsIDHandler,
```
it makes sense to set them.

peterjan (approved):
We should probably create an issue to add benchmark for some of the most important `explored` database queries.

ChrisSchinnerl (approved):

Issue #155, opened 2025-01-03
Refactor Tests


n8maninger:
When writing tests it’s important to focus on testing the smallest, most isolated piece of functionality you can. This keeps tests focused and makes debugging future regressions easier.

Take the sqlite package as an example. The goal for the tests in this package should be to test how SQL queries interact with the database. Using `TestScan` as an example: it spins up a whole explorer node and brings in a chain manager, syncer, and all the other unrelated dependencies. Instead, aim to mock the specific database state needed for each test. It can be annoying and take more effort to set up, but it keeps your tests clean and focused. The goal is to test SQL interactions -- not the whole scan process.

If you want to test that scanning is working as expected, that should happen where the scanning code lives -- in the explorer package. The goal of tests isn’t just to say "this works," it's also to make breaking logic changes easier to detect. That way we've got multiple levels of validation that scanning is working as expected.

Glancing through the code, I would recommend creating sub-issues and doing this piece by piece.

chris124567:
Agree.

With regards to the consensus tests, I tried to keep the v2 tests more "focused" in this regard but there's definitely improvements to be made across the board.  But part of the reason we don't mock database state is because we are testing the functions that retrieve from the database but we also want to make sure that it was inserted properly.  If we just have a test that inserts the data (i.e. transactions with file contracts) then we're merely testing whether the insert queries succeeded or not and not whether the queries themselves were correct or not.

n8maninger:
Using your example of a transaction with file contract, I would structure one of the sqlite tests like this:

1. Build an apply update that includes a transaction and a file contract
2. Use getter to ensure the contract and transaction were both added as expected and return the expected state
3. Build an apply update that includes a transaction and a revision
4. Use the getter again to ensure the contract was updated and the transaction exists.

In this theoretical test, the goal isn't necessarily to only test that the insert query works (although I would recommend a separate test for that too), but to test the expected lifecycle and ensure database consistency. Most importantly, the focus is on verifying the state of the database and avoids dependencies like the chain manager or indexing loop which could unintentionally alter the state. Additional tests could expand on this further, such as including a storage proof, a revert update, etc. 

Then in the explorer package, I would recommend a separate suite of tests to validate that contracts from the blockchain are being indexed correctly. For example:

1. Build a transaction that includes a file contract
2. Mine the block
3. Use the explorer getter methods to ensure the explorer is returning the expected state
4. Build a transaction that includes a revision for the contract
5. Mine the block
6. Check the state again

[...]

PR #204, opened 2025-04-18
Refactor tests part 1

chris124567:
The current set of tests work and provide good coverage but as noted in #155 have become unwieldy - they started out simple but as new features were added to explored they were extended instead of adding separate tests.  They also include things like the `chain.Manager` that we are not interested in testing here.  This PR includes more focused tests for:

- Tip
- Block
-  BestTip
- Transactions
- TransactionChainIndices
- Balance
- SiacoinElements (including ephemeral outputs)
- SiafundElements (including ephemeral outputs)

In the process of writing these tests I discovered and fixed two bugs:
1. If, when the first time we encounter a transaction, there is an identical transaction in the same block, its fields would not be inserted.  This was fixed by adding adding checks to the loop in `addTransactions`.
2. If we reverted a block and it contained a transaction not used in any other block, some state about the transaction would linger in the `transactions` table though all references to it in `block_transactions` would be removed.  This was resolved by adding a trigger upon deletion in block_transactions to remove these "orphaned" transactions.

Once we reach coverage parity with the old tests we should be able to remove them, but not yet.


persist/sqlite/consensus.go
```diff
@@ -995,6 +1011,42 @@ func (ut *updateTx) Metrics(height uint64) (explorer.Metrics, error) {
...
+		state.SpentSiafundElements,
+		append(state.NewSiafundElements, state.EphemeralSiafundElements...),
+	); err != nil {
+		return fmt.Errorf("RevertIndex: failed to update siafund output state: %w", err)
+	} else if err := updateBalances(ut.tx, state.Metrics.Index.Height, state.SpentSiacoinElements, state.NewSiacoinElements, state.SpentSiafundElements, state.NewSiafundElements); err != nil {
+		return fmt.Errorf("RevertIndex: failed to update balances: %w", err)
+	} else if _, err := updateFileContractElements(ut.tx, true, state.Metrics.Index, state.Block, state.FileContractElements); err != nil {
+		return fmt.Errorf("RevertIndex: failed to update file contract state: %w", err)
+	} else if _, err := updateV2FileContractElements(ut.tx, true, state.Metrics.Index, state.Block, state.V2FileContractElements); err != nil {
+		return fmt.Errorf("ApplyIndex: failed to add v2 file contracts: %w", err)
```
peterjan, line 1036:
```suggestion
		return fmt.Errorf("RevertIndex: failed to update v2 file contracts: %w", err)
```

persist/sqlite/init.sql
```diff
@@ -278,6 +291,19 @@ CREATE TABLE v2_block_transactions (
 CREATE INDEX v2_block_transactions_block_id_index ON v2_block_transactions(block_id);
 CREATE INDEX v2_block_transactions_transaction_id_block_id ON v2_block_transactions(transaction_id, block_id);
 
+CREATE TRIGGER delete_orphan_v2_transaction
```
peterjan, line 294: Little of PTSD when it comes to using triggers 😅 . Could you elaborate why the database schema can't be a chain of entities that depend upon each other, and eventually a block that can be referenced and have a `CASCADE DELETE` when we revert the chain index at a certain height? We can probably have `v2_transactions` reference `v2_block_transactions` instead of the other way around and add a `CASCADE DELETE` there?

persist/sqlite/consensus_refactored_test.go
```diff
@@ -0,0 +1,943 @@
...
+func newStoreRefactored(t *testing.T, v2 bool, f func(*consensus.Network, types.Block)) (types.Block, consensus.State, *chain.DBStore, explorer.Store) {
+	log := zaptest.NewLogger(t)
+	dir := t.TempDir()
+
+	db, err := sqlite.OpenDatabase(filepath.Join(dir, "explored.sqlite3"), log.Named("sqlite3"))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	bdb, err := coreutils.OpenBoltChainDB(filepath.Join(dir, "consensus.db"))
```
ChrisSchinnerl, line 30: Use `chain.NewMemDB` for testing

persist/sqlite/consensus_refactored_test.go
```diff
@@ -0,0 +1,943 @@
...
+
+	cs, au := consensus.ApplyBlock(cs, b, bs, time.Time{})
+	return cs, chain.ApplyUpdate{
+		ApplyUpdate: au,
+		Block:       b,
+		State:       cs,
+	}
+}
+
+func revertUpdate(t *testing.T, prevState consensus.State, bs consensus.V1BlockSupplement, b types.Block) chain.RevertUpdate {
```
ChrisSchinnerl, line 79: If you make this and `applyUpdate` methods on a helper type, they should only need the `t` and `types.Block` as their arguments and could be `ApplyBlock` and `RevertBlock`. The current state could be stored on the helper type and the supplement computed within the helper using the store.

If you are feeling extra fancy you can even keep a slice of blocks on your helper type so that you don't need to pass a block to `RevertBlock` to move some complexity that is currently in reorg tests to the helper code.

e.g.

```go
foo.ApplyBlock(t, b) // apply some custom block
foo.MineBlocks(t, 10) // mine blocks on top of b

for i := 0; i < 10; i++ {
  foo.RevertBlock() // revert the latest block
}

assertTip() // should be back to initial block
```

Of course this is just a very simple example, but it could make reorg tests very clean. The goal for tests should be to really make the test about the behavior you want to test rather than distracting glue code.

persist/sqlite/consensus_refactored_test.go
```diff
@@ -0,0 +1,943 @@
...
+
+func TestSiacoinOutput(t *testing.T) {
+	pk1 := types.GeneratePrivateKey()
+	addr1 := types.StandardUnlockHash(pk1.PublicKey())
+
+	pk2 := types.GeneratePrivateKey()
+	addr2 := types.StandardUnlockHash(pk2.PublicKey())
+
+	genesisBlock, cs, store, db := newStoreRefactored(t, false, func(network *consensus.Network, genesisBlock types.Block) {
+		genesisBlock.Transactions[0].SiacoinOutputs[0].Address = addr1
```
ChrisSchinnerl, line 96: The callback isn't great. I'd either double-down on the functional argument pattern and add options like `WithGenesisPayoutAddr` or pass the `network` and `genesisBlock` in as regular arguments. You could then have separate helpers like `v1TestNetwork(payoutAddr types.Address)` that you use to create the block and network whenever you need a specific payout address.

persist/sqlite/consensus_refactored_test.go
```diff
@@ -0,0 +1,943 @@
...
+	testutil.SignTransaction(cs, pk1, &txn1)
+
+	prevState := cs
+	b := testutil.MineBlock(cs, []types.Transaction{txn1}, types.VoidAddress)
+	cs, au := applyUpdate(t, cs, store.SupplementTipBlock(b), b)
+	if err := db.UpdateChainState(nil, []chain.ApplyUpdate{au}); err != nil {
+		t.Fatal(err)
+	}
+
+	{
```
ChrisSchinnerl, line 120: Avoid this pattern with the indentations and prefer an inline helper with a quick comment to explain what it does and then a quick comment per line what you are testing in particular.

```go
// helper to assert the Siacoin element in the db has the right source, index and output
assertElement := func(scoid types.SiacoinOutputID, index *types.ChainIndex, sco types.SiacoinOutput) {
        t.Helper()
	sces, err := db.SiacoinElements([]types.SiacoinOutputID{scID})
	if err != nil {
		t.Fatal(err)
	}
	testutil.Equal(t, "len(sces)", 1, len(sces))

	sce := sces[0]
	testutil.Equal(t, "sce.Source", explorer.SourceTransaction, sce.Source)
	testutil.Equal(t, "sce.SpentIndex", index, *sce.SpentIndex)
	testutil.Equal(t, "sce.SiacoinElement.SiacoinOutput", sco, sce.SiacoinOutput)
}

// quick comment on why this is the genesis output
assertElement(scID, cs.Index, genesisBlock.Transactions[0].SiacoinOutputs[0])

// TODO: next call and comment about why things are the way they are and so on
```

That reduces the lines of code in your test but also increases readability. Tests should be reasonably well documented to make sure anyone on the team can understand what is being tested immediately and why certain assumptions are made if that's not obvious from the code.

If the exact same code is reused across tests you might as well make it a helper function in your test package rather than inlining it. But only if it makes sense. I wouldn't force it at the cost of less test coverage for individual tests.

persist/sqlite/consensus_refactored_test.go
```diff
@@ -0,0 +1,943 @@
...
+		if err != nil {
+			t.Fatal(err)
+		}
+		testutil.Equal(t, "len(txns)", 0, len(txns))
+	}
+}
+
+func TestEphemeralSiacoinOutput(t *testing.T) {
+	pk1 := types.GeneratePrivateKey()
+	addr1 := types.StandardUnlockHash(pk1.PublicKey())
```
ChrisSchinnerl, line 193: (nit) since you need the unlock conditions anyway I'd create them up here as well and go from private key -> unlock conditions -> unlock hash to avoid passing the public key twice

```go
pk1 := types.GeneratePrivateKey()
uc1 := types.StandardUnlockConditions(pk1.PublicKey())
addr1 := uc1.UnlockHash()
```
you can also avoid `addr1` and just inline the `uc1.UnlockHash` call.

ChrisSchinnerl (changes requested):
Before further diving into this I'd break out the bug fixes into separate PRs and create a PR for the `TestSiacoinOutput` and `TestEphemeralSiacoinOutput` tests which we try to get as clean as possible before moving on to the next couple of tests.

chris124567:
All of these implemented in other PRs

PR #207, opened 2025-04-23
Fix bug where state about some transactions remained even after revert

chris124567:
If we reverted a block and it contained a transaction not used in any other block, some state about the transaction would linger in the transactions/v2_transactions table though all references to it in `block_transactions` would be removed. Fixed by adding functions we call only in `RevertIndex` to remove these "orphaned" transactions.


persist/sqlite/revert.go
```diff
@@ -0,0 +1,92 @@
...
+	"fmt"
+
+	"go.sia.tech/core/types"
+	"go.sia.tech/explored/explorer"
+)
+
+// deleteV1Transactions deletes the transactions from the database if they are
+// not referenced in any blocks.
+func deleteV1Transactions(tx *txn, txns []types.Transaction) error {
+	stmt, err := tx.Prepare(`DELETE FROM transactions AS t
```
peterjan, line 13: defer close

persist/sqlite/consensus_test.go
```diff
@@ -2509,3 +2509,48 @@ func TestMetricCirculatingSupply(t *testing.T) {
...
+		txns, err := db.Transactions([]types.TransactionID{txn.ID()})
+		if err != nil {
+			t.Fatal(err)
+		}
+		testutil.Equal(t, "len(txns)", 1, len(txns))
+		testutil.CheckTransaction(t, txn, txns[0])
+	}
+
+	{
+		// mine to trigger a reorg
```
peterjan, line 2535: could be a helper

ChrisSchinnerl: Same thing further down in the diff. We could use a helper that starts mining and applying blocks starting from a certain state.
Ideally this whole block would be a single line of code. The previous and next blocks could also be a helper that you declare inline. Then you'd end up with something like

```go
assertTransactions([]types.Transaction{txn})
mineBlocks(genesisState, 2)
assertTransactions([]types.Transacdtion{}) 
```

ChrisSchinnerl (changes requested):

chris124567:
I think it makes more sense to make a better refactored test in a subsequent PR with test helpers like those in https://github.com/SiaFoundation/explored/pull/209, otherwise we'll have a create a bunch of helpers that we don't end up using in the long run.

persist/sqlite/v2transactions.go
```diff
@@ -113,7 +114,9 @@ func getV2TransactionBase(tx *txn, txnIDs []types.TransactionID) ([]int64, []exp
 	for _, id := range txnIDs {
 		var txn explorer.V2Transaction
 		var newFoundationAddress types.Address
-		if err := stmt.QueryRow(encode(id)).Scan(&dbID, decode(&txn.ID), decodeNull(&newFoundationAddress), decode(&txn.MinerFee), &txn.ArbitraryData); err != nil {
+		if err := stmt.QueryRow(encode(id)).Scan(&dbID, decode(&txn.ID), decodeNull(&newFoundationAddress), decode(&txn.MinerFee), &txn.ArbitraryData); errors.Is(err, sql.ErrNoRows) {
```
peterjan, line 117: unrelated to the PR? could add missing unit test for this

peterjan (approved):

ChrisSchinnerl (approved):

persist/sqlite/v2transactions.go
chris124567: It's already covered by the final check in TestV2TransactionChainIndices

PR #209, opened 2025-04-23
Add refactored TestSiacoinOutput and TestEphemeralSiacoinOutput

chris124567:
Incorporate ideas from review in #204


persist/sqlite/consensus_refactored_test.go
```diff
@@ -0,0 +1,275 @@
...
+	"go.sia.tech/core/types"
+	"go.sia.tech/coreutils/chain"
+	ctestutil "go.sia.tech/coreutils/testutil"
+	"go.sia.tech/explored/explorer"
+	"go.sia.tech/explored/internal/testutil"
+	"go.sia.tech/explored/persist/sqlite"
+	"go.uber.org/zap/zaptest"
+)
+
+type node struct {
```
peterjan, line 19: `testChain` and `newTestChain`?

persist/sqlite/consensus_refactored_test.go
```diff
@@ -0,0 +1,275 @@
...
+	db    explorer.Store
+	store *chain.DBStore
+
+	blocks []types.Block
+	states []consensus.State
+}
+
+func newStoreRefactored(t *testing.T, v2 bool, f func(*consensus.Network, types.Block)) *node {
```
peterjan, line 27: I'd probably do either:

- `newTestChainV1`
- `newTestChainV2`

or (my preference I think)

`newTestChain(t *testing, network *consensus.Network, genesis types.Block)`
I would get rid of 'f', at the very least give it a more descriptive name.

persist/sqlite/consensus_refactored_test.go
```diff
@@ -0,0 +1,275 @@
...
+	_, au := consensus.ApplyBlock(network.GenesisState(), genesisBlock, bs, time.Time{})
+	if err := db.UpdateChainState(nil, []chain.ApplyUpdate{{
+		ApplyUpdate: au,
+		Block:       genesisBlock,
+		State:       genesisState,
+	}}); err != nil {
+		t.Fatal(err)
+	}
+
+	t.Cleanup(func() {
```
peterjan, line 62: `nit` call `t.Cleanup(func() {db.Close()})` immediately after opening the database

persist/sqlite/consensus_refactored_test.go
```diff
@@ -0,0 +1,275 @@
...
+			t.Fatal(err)
+		}
+		testutil.Equal(t, "len(txns)", 1, len(txns))
+
+		testutil.CheckTransaction(t, txn, txns[0])
+	}
+}
+
+// helper to assert the Siacoin element in the db has the right source, index and output
+func (n *node) assertSCE(t *testing.T, scID types.SiacoinOutputID, index *types.ChainIndex, sco types.SiacoinOutput) {
```
peterjan, line 137: Most of the time we want to figure out whether an output was spent or not, so I'd go for  `isSpent(id types.SiacoinOutputID) bool` .If you pass in the genesis block and network you also avoid the slightly awkward `n.genesis().Transactions[0].SiacoinOutputID(0)` all over the place. I'd also add a `mineTransaction` perhaps. It should be closer to the following, but even that is still quite hard to read

```go
        network, genesis := ctestutil.Network()
	genesis.Transactions[0].SiacoinOutputs[0].Address = addr1
	chain := newTestChain(t, network, genesis)

	// genesis output should be unspent
	scID := genesis.Transactions[0].SiacoinOutputID(0)
	if chain.isSpent(t, scID) {
		t.Fatal("genesis output should not be spent")
	}

	// spend the output
	txn1 := types.Transaction{
		SiacoinInputs: []types.SiacoinInput{{
			ParentID:         scID,
			UnlockConditions: uc1,
		}},
		SiacoinOutputs: []types.SiacoinOutput{{
			Address: addr2,
			Value:   genesis.Transactions[0].SiacoinOutputs[0].Value,
		}},
	}
	testutil.SignTransaction(chain.tipState(), pk1, &txn1)
	chain.mineTransaction(t, testutil.MineBlock, txn1)

	// assert output was unspent
	if !chain.isSpent(t, scID) {
		t.Fatal("genesis output should be spent")
	}
```

peterjan (commented):
Definitely cleaner. Although I was expecting to see a side-by-side, but I guess these tests are newly introduced? The ultimate goal would be to really see how a test used to look and how it looks now, then we can clearly see what we really need and write good test helpers/structs for it.

chris124567:
These test the same things as the tests in #204, except TestSiacoinOutput doesn't check the transactions before/after revert because #207 isn't merged yet.

persist/sqlite/consensus_refactored_test.go
chris124567: I feel like there's not much to lose from checking the other fields of the siacoin element.  The spent status is the only thing that is *supposed* to change after block application/revert but if there's a bug then other things could too.

peterjan (approved):

ChrisSchinnerl (approved):

PR #212, opened 2025-04-24
Fix chain migration issue

chris124567:
Fix crash stemming from chain migrations.  We still have to reset all the state in the database, but at least this doesn't panic.

Similar solution used in walletd: https://github.com/SiaFoundation/walletd/blob/d91147b7ec52b2261779949905c8e0fd3c0ad5c7/wallet/manager.go#L727


explorer/explorer.go
```diff
@@ -112,19 +114,24 @@ type Explorer struct {
 
 func (e *Explorer) syncStore(index types.ChainIndex, batchSize int) error {
 	for index != e.cm.Tip() {
-		crus, caus, err := e.cm.UpdatesSince(index, batchSize)
-		if err != nil {
-			return fmt.Errorf("failed to subscribe to chain manager: %w", err)
-		}
+		select {
```
peterjan, line 117:
```go
for index != e.cm.Tip() && e.ctx.Err() == nil {
    ...
}
return e.ctx.Err()
```
is a little bit cleaner

explorer/explorer.go
```diff
@@ -172,7 +179,19 @@ func NewExplorer(cm ChainManager, store Store, indexCfg config.Index, scanCfg co
...
+				switch {
+				case errors.Is(err, context.Canceled):
+					return
+				case strings.Contains(err.Error(), "missing block at index"):
+					log.Warn("missing block at index, resetting chain state", zap.Stringer("id", lastTip.ID), zap.Uint64("height", lastTip.Height))
+					if err := e.s.ResetChainState(); err != nil {
+						log.Panic("failed to reset explorer state", zap.Error(err))
+					}
+					// trigger resync
+					reorgChan <- types.ChainIndex{}
```
peterjan, line 191: this potentially deadlocks the loop I believe, should do select with a default case to ensure it doesn't block on this

explorer/explorer.go
```diff
@@ -172,7 +179,19 @@ func NewExplorer(cm ChainManager, store Store, indexCfg config.Index, scanCfg co
 				e.log.Error("failed to get tip", zap.Error(err))
 			}
 			if err := e.syncStore(lastTip, indexCfg.BatchSize); err != nil {
-				e.log.Panic("failed to sync store", zap.Error(err))
+				switch {
+				case errors.Is(err, context.Canceled):
+					return
+				case strings.Contains(err.Error(), "missing block at index"):
```
peterjan, line 185: we should open a PR on coreutils to expose this error, it only takes a second and we're updating coreutils/all dependencies every day anyway

explorer/explorer.go
ChrisSchinnerl: I honestly don't like that pattern. I'd rather have the logic for interrupting the loop isolated in a `select` and then return `nil` to have clear success and interrupt paths rather than mixing them.

explorer/explorer.go
ChrisSchinnerl: I'd create a separate issue to track this. This shouldn't block this PR and we have multiple repos that need to be updated once we have it.

explorer/explorer_test.go
```diff
@@ -0,0 +1,78 @@
...
+		t.Fatal(err)
+	}
+	cm := chain.NewManager(store, genesisState)
+
+	db, err := sqlite.OpenDatabase(filepath.Join(dir, "explored.sqlite3"), log.Named("sqlite3"))
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer db.Close()
+
```
ChrisSchinnerl, line 37: Running the test like this means the tables are mostly empty. There are no transactions with siacoins, siafunds, contracts, resolutions etc. It would be good to have some basic traffic here to make sure the database is filled.
Then we corrupt it, resync and check the tip + a few of the things we added such as transactions.

persist/sqlite/consensus.go
```diff
@@ -1204,3 +1204,17 @@ func (s *Store) UpdateChainState(reverted []chain.RevertUpdate, applied []chain.
 		return nil
 	})
 }
+
+// ResetChainState implements explorer.Store
+func (s *Store) ResetChainState() error {
+	return s.transaction(func(tx *txn) error {
+		tables := []string{"blocks", "transactions", "v2_transactions", "host_info", "address_balance", "state_tree"}
```
ChrisSchinnerl, line 1211: I'd spin up a mainnet explorer, let it sync and then manually run this query on these tables. Then check the database and see if it's in the state you'd expect. To make sure it works on a mature database.

explorer/explorer.go
chris124567: https://github.com/SiaFoundation/coreutils/pull/217

chris124567: https://github.com/SiaFoundation/coreutils/pull/217

persist/sqlite/consensus.go
chris124567: I tried on nearly fully synced mainnet but it was taking so long that I cancelled before it finished.  In that case we might just have to drop all the tables and recreate them.

chris124567: Ok tested with new logic performance is good now

persist/sqlite/consensus.go
```diff
@@ -1204,3 +1203,50 @@ func (s *Store) UpdateChainState(reverted []chain.RevertUpdate, applied []chain.
 		return nil
 	})
 }
+
+// ResetChainState implements explorer.Store
+func (s *Store) ResetChainState() error {
+	if _, err := s.db.Exec(`PRAGMA foreign_keys = OFF;`); err != nil {
```
ChrisSchinnerl, line 1209: Instead of this use `PRAGMA defer_foreign_keys=ON` within the transaction similar to how `hostd` does it in `upgradeDatabase` for example.

ChrisSchinnerl (changes requested):

persist/sqlite/consensus.go
chris124567: I get this error when I try this:

```
failed to reset explorer state  {"error": "ResetChainState: failed to delete and reinit database: failed to drop table siacoin_elements: no such table: main.blocks"}
```
Though based on the [documentation](https://www.sqlite.org/pragma.html#pragma_defer_foreign_keys) it seems like it should work.

explorer/explorer_test.go
chris124567: If we add contract formation/resolution code the test will end up being much larger than it currently is.  Especially because we don't have all the helpers we have in the `sqlite` package.  The genesis block has SC/SF output transactions in it at least so we could check those.

persist/sqlite/consensus.go
```diff
@@ -1204,3 +1203,50 @@ func (s *Store) UpdateChainState(reverted []chain.RevertUpdate, applied []chain.
...
+
+		var names []string
+		for rows.Next() {
+			var name string
+			if err := rows.Scan(&name); err != nil {
+				return fmt.Errorf("failed to scan table name: %w", err)
+			}
+			names = append(names, name)
+		}
+		rows.Close()
```
peterjan, line 1226: defer the `Close`, we're not closing it when we fail to scan the table name

persist/sqlite/consensus.go
peterjan: Considering the implementation of `ResetChainState`, I don't see the benefit of using `defer_foreign_keys`, right? It's very different from `hostd`'s ``upgradeDatabase`. I think if we really want to do it we'd have to be able to control the order with which we drop the tables.

ChrisSchinnerl: @chris124567 what about moving the `foreign_keys = OFF` into the transaction, the `foreign_keys = ON` too and then at the very end a `foreign_key_check`?

The check itself should be super fast on an empty database and guarantee that we dropped all tables even if the migration isn't very spectacular itself.

chris124567: I get the same error with that:

```
		if _, err := tx.Exec(`PRAGMA foreign_keys = OFF;`); err != nil {
			return fmt.Errorf("failed to disable foreign key constraints: %w", err)
		}

		rows, err := tx.Query(`SELECT name FROM sqlite_master WHERE type='table' AND name NOT LIKE 'sqlite_%'`)
		if err != nil {
			return fmt.Errorf("failed to get list of tables")
		}
		defer rows.Close()

		var names []string
		for rows.Next() {
			var name string
			if err := rows.Scan(&name); err != nil {
				return fmt.Errorf("failed to scan table name: %w", err)
			}
			names = append(names, name)
		}

		for _, name := range names {
			if _, err := tx.Exec(fmt.Sprintf(`DROP TABLE IF EXISTS %s`, name)); err != nil {
				return fmt.Errorf("failed to drop table %s: %w", name, err)
			}
		}

		target := int64(len(migrations) + 1)
		if err := s.initNewDatabase(tx, target); err != nil {
			return fmt.Errorf("failed to reinit tables: %w", err)
		}

		if _, err := tx.Exec(`PRAGMA foreign_keys = ON;`); err != nil {
			return fmt.Errorf("failed to enabale foreign key constraints: %w", err)
		}
		if _, err := tx.Exec(`foreign_key_check`); err != nil {
			return fmt.Errorf("failed to enabale foreign key constraints: %w", err)
		}
```

```
{"error": "ResetChainState: failed to delete and reinit database: failed to drop table siacoin_elements: no such table: main.blocks"}
```

chris124567: Alright, I think the most reliable thing to do here is to drop the dependent tables first.  I don't expect we'll be adding new tables any time soon, at most new fields to existing tables.

persist/sqlite/consensus.go
```diff
@@ -1204,3 +1203,80 @@ func (s *Store) UpdateChainState(reverted []chain.RevertUpdate, applied []chain.
 		return nil
 	})
 }
+
+// ResetChainState implements explorer.Store
+func (s *Store) ResetChainState() error {
+	if err := s.transaction(func(tx *txn) error {
+		if _, err := tx.Exec(`PRAGMA defer_foreign_keys=ON`); err != nil {
```
peterjan, line 1210: don't forget to turn it off

ChrisSchinnerl: It's a `defer` pragma now so it will not affect the foreign key checks beyond this transaction

ChrisSchinnerl (approved):

ChrisSchinnerl (approved):

PR #216, opened 2025-04-30
Add refactored balance tests

chris124567:
- Add siafund versions of `TestSiacoinOutput`/`TestEphemeralSiacoinOutput`
- Add tests of balances for SC/SF including with reverts and transactions with ephemeral outputs
- Remove existing `TestSiafundBalance` as redundant - I thought it would be possible to remove more tests but some of the old tests that test balances also test `UnspentSiacoinOutputs`/`UnspentSiafundOutputs`.  I will soon make a PR with tests for those functions.  Once those are it gone, it should be possible to remove `TestBalance` and `TestRevertBalance`.


peterjan (approved):
lgtm 👍

ChrisSchinnerl (approved):

PR #219, opened 2025-05-03
Add new refactored tests

chris124567:
- Add tests for `UnspentSiacoinOutputs` and `UnspentSiafundOutputs`
- Add tests for `Block` (v1 and v2)
- Add test where we wait for output to mature and check immature and matured balance
- Add test for retrieving events (only payout event)

Remove old tests:
- `TestBalance`
- `TestRevertBalance`


persist/sqlite/consensus_refactored_test.go
```diff
@@ -570,6 +600,78 @@ func TestV2TransactionChainIndices(t *testing.T) {
...
+		checkBalance(addr1, types.ZeroCurrency, val)
+		n.mineTransactions(t)
+	}
+
+	checkBalance(addr1, val, types.ZeroCurrency)
+
+	for range n.network.MaturityDelay {
+		n.revertBlock(t)
+		checkBalance(addr1, types.ZeroCurrency, val)
+	}
```
peterjan, line 639: we can add a final check `checkBalance(addr1, types.ZeroCurrency, types.ZeroCurrency)`

persist/sqlite/consensus_refactored_test.go
```diff
@@ -570,6 +600,78 @@ func TestV2TransactionChainIndices(t *testing.T) {
...
+		testutil.Equal(t, "events", 1, len(events))
+
+		ev0 := events[0].Data.(explorer.EventPayout)
+		testutil.Equal(t, "event 0 output ID", n.tipState().Index.ID.MinerOutputID(0), ev0.SiacoinElement.ID)
+		testutil.Equal(t, "event 0 output source", explorer.SourceMinerPayout, ev0.SiacoinElement.Source)
+	}
+
+	n.revertBlock(t)
+
+	{
```
peterjan, line 666: still don't quite get the code blocks but these tests are very readable and easy to follow now 👍

persist/sqlite/consensus_refactored_test.go
```diff
@@ -909,3 +1011,304 @@ func TestMissingBlock(t *testing.T) {
...
+		SiafundElement: types.SiafundElement{
+			ID:            genesisOutputID,
+			SiafundOutput: genesisOutput,
+		},
+	})
+	checkSiafundOutputs(addr2)
+	checkSiafundOutputs(addr3)
+}
+
+func TestBlock(t *testing.T) {
```
peterjan, line 1209: What's the intention of this test? There's no solid assertions here. We would probably have to assert counts at the very least because we're just ranging over the blocks and seeing if they are present in the database, but those transactions don't really matter.

persist/sqlite/consensus_refactored_test.go
```diff
@@ -909,3 +1011,304 @@ func TestMissingBlock(t *testing.T) {
...
+	n.mineTransactions(t, txn1, txn2, txn3)
+
+	checkBlocks()
+
+	n.revertBlock(t)
+
+	checkBlocks()
+}
+
+func TestV2Block(t *testing.T) {
```
peterjan, line 1262: same here

peterjan (commented):
good job on the helpers 👍 it's a lot more fun reviewing these tests in the new framework

persist/sqlite/consensus_refactored_test.go
chris124567: To check that nonce/timestamp/miner payouts are set.  I simplified it a lot in recent commit should be better now.

persist/sqlite/consensus_refactored_test.go
chris124567: I guess we could remove them, it's mostly so I don't have to use different variables names and to logically isolate each set of check since tend to be a couple lines long.  Ideally we'd have functions like `assertTransactionsNotExist` or `assertNoEvents` but it doesn't really seem right if we only use it once.

persist/sqlite/consensus_refactored_test.go
peterjan: Perhaps we should pass an expected amount of blocks? We could do `checkBlocks(1) - mine - checkBlocks(2) - revert - checkBlocks(1)` perhaps? Fine to resolve if that doesn't make sense, my issue with the test is that `checkBlocks` just assumes `n.blocks` is being updated. Obviously that is the case but this test would still pass if we wouldn't append or remove from `n.blocks`

chris124567:
Rebased and also removed TestSendTransactions since we separately cover everything in it.

PR #220, opened 2025-05-04
use threadgroup for shutdowns

n8maninger:
Removes the context and context cancel func on the explorer struct in favor of the more idiomatic threadgroup.

### DIFF
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -1,0 +1,26 @@
+on:
+  push:
+    branches: [master]
+
+permissions:
+  contents: write
+  pull-requests: write
+
+name: Create Release PR
+jobs:
+  prepare-release:
+    if: "!contains(github.event.head_commit.message, 'chore: prepare release')" # Skip merges from releases
+    runs-on: ubuntu-latest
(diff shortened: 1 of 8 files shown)


.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,26 @@
+on:
```
peterjan, line 1: did you mean to commit this and the `knope.toml` in this PR?

explorer/explorer.go
```diff
@@ -211,16 +212,13 @@ func NewExplorer(cm ChainManager, store Store, indexCfg config.Index, scanCfg co
 
 // Shutdown tries to close the scanning goroutines in the explorer.
 func (e *Explorer) Shutdown(ctx context.Context) error {
-	e.ctxCancel()
-	e.locator.Close()
-
-	done := make(chan struct{})
+	done := make(chan struct{}, 1)
```
peterjan, line 215: shouldn't be buffered

.github/workflows/prepare-release.yml
n8maninger: I did

PR #224, opened 2025-05-07
Fix file contract revert resolved state bug

chris124567:
- Fix bug where resolved/valid state were not reset after revert.  The code in `updateFileContractElements` failed to account for the fact that if we are reverting and the contract is resolved or valid in the file contract element diff from core, then we should do the opposite and set those to false.
- Move resolved/valid state to the `last_contract_revision` table from `file_contract_elements`.  Before, when we retrieved a transaction that created or revised a file contract, in the enhanced `explorer.ExtendedFileContract` struct the resolved/valid fields would reflect whether it was resolved/valid at the time of that transaction.  This is not really useful, so it makes more sense to have those values reflect the latest state.  This is also in line with how we handle resolution state for v2 contracts.
- Add basic tests for file contracts

### DIFF
--- a/persist/sqlite/consensus_refactored_test.go
+++ b/persist/sqlite/consensus_refactored_test.go
@@ -255,6 +255,9 @@ func (n *testChain) assertFCE(t *testing.T, fcID types.FileContractID, expected 
 	testutil.Equal(t, "ExtendedFileContract", expected, fce)
 }
 
+// assertTransactionContracts asserts that the enhanced FileContracts
+// (revisions = false) or FileContractRevisions (revisions = true) in a
+// transaction retrieved from the explorer match the expected contracts.
 func (n *testChain) assertTransactionContracts(t *testing.T, txnID types.TransactionID, revisions bool, expected ...explorer.ExtendedFileContract) {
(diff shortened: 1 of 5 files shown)


peterjan (dismissed):

persist/sqlite/consensus_refactored_test.go
```diff
@@ -237,6 +237,59 @@ func (n *testChain) assertSFE(t *testing.T, sfID types.SiafundOutputID, index *t
...
+	testutil.Equal(t, "len(fces)", 1, len(fces))
+
+	fce := fces[0]
+	// We aren't trying to compare a core type with an explorer type so we can
+	// just directly compare.  If they are not equal a diff with field names will
+	// be printed.
+	testutil.Equal(t, "ExtendedFileContract", expected, fce)
+}
+
+func (n *testChain) assertTransactionContracts(t *testing.T, txnID types.TransactionID, revisions bool, expected ...explorer.ExtendedFileContract) {
```
peterjan, line 258: `nit` add docstring

ChrisSchinnerl (approved):

PR #225, opened 2025-05-07
Add refactored file contract revision tests

chris124567:
- Add refactored file contract revision tests, including revising same contract multiple times in one block
- Move refactored v2 tests to separate file
- Remove old `TestSameBlock`/`TestBlockSameV2Transaction` because we already cover the case of same transaction in one block in `TestTransactionChainIndices`/`TestV2TransactionChainIndices`


persist/sqlite/consensus_refactored_test.go
```diff
@@ -1556,10 +1460,246 @@ func TestFileContractMissed(t *testing.T) {
 		testutil.Equal(t, "len(fces)", 0, len(fces))
 	}
 
+	n.assertContractRevisions(t, fce.ID)
+}
+
+func signRevisions(cs consensus.State, pk types.PrivateKey, txn *types.Transaction) {
+	appendSig := func(key types.PrivateKey, pubkeyIndex uint64, parentID types.Hash256) {
```
peterjan, line 1467: inline the helper, only used once

peterjan (approved):
LGTM

ChrisSchinnerl (approved):

persist/sqlite/consensus_refactored_test.go
chris124567: It's used in TestFileContractRevision and TestFileContractMultipleRevisions

PR #229, opened 2025-05-13
Add test for ContractsKey and Metrics

chris124567:
Add refactored:
- `TestMetrics`
- `TestFileContractsKey`

Remove old:
- `TestEphemeralFileContract`
- `TestMetricCirculatingSupply`

Now that we have metrics done, pretty much the only thing holding up the removal of most of the old tests is events.


persist/sqlite/consensus_refactored_test.go
```diff
@@ -1703,3 +1713,248 @@ func TestFileContractMultipleRevisions(t *testing.T) {
...
+		}},
+		SiacoinOutputs: []types.SiacoinOutput{{
+			Address: addr1,
+			Value:   val.Sub(fc.Payout.Mul64(2)),
+		}},
+		FileContracts: []types.FileContract{fc, fc},
+	}
+	testutil.SignTransaction(n.tipState(), pk1, &txn1)
+
+	if subsidy, ok := n.tipState().FoundationSubsidy(); ok {
```
peterjan, line 1863: `nit` I'd mine an empty block after asserting the circulating supply, that way testing that funds locked away in contract impact the circulating supply is slightly cleaner

persist/sqlite/consensus_refactored_test.go
```diff
@@ -1703,3 +1713,248 @@ func TestFileContractMultipleRevisions(t *testing.T) {
 
 	n.assertContractRevisions(t, fce.ID)
 }
+
+func TestFileContractsKey(t *testing.T) {
```
peterjan, line 1717: Does `TestFileContractsKey` cover everything we used to assert? It seems we're missing a couple of things like `assertContractRevisions`

peterjan (approved):

PR #231, opened 2025-05-15
Do not use embedded marshalers for SiacoinInput and SiafundInput

chris124567:
https://github.com/SiaFoundation/core/commit/1b92425edab8ee7ad7b746ae88a281da9023c020 recently added custom marshalers to SiacoinInput and SiafundInput.  We embedded these types in `explorer.SiacoinInput` and `explorer.SiafundInput`, so those marshalers were used instead of Go's default encoder for structs, which resulted in the `value` field not being included in responses.  This PR adds custom marshalers to those types to prevent the embedded type from overriding them.  This was actually caught by the tests in the `api` module, but we were using a version of `core` from 2 weeks ago, so this bug only appeared in clusterd settings where core directly from master was being used.


explorer/types.go
```diff
@@ -596,3 +596,35 @@ type HostQuery struct {
 	AcceptContracts      *bool             `json:"acceptContracts,omitempty"`
 	Online               *bool             `json:"online,omitempty"`
 }
+
+// MarshalJSON implements json.Marshaler.  The embedded types.SiacoinInput
+// in our SiacoinInput has its own marshaler that will override default
+// marshaling and result in fields we expect being missing.
+func (e SiacoinInput) MarshalJSON() ([]byte, error) {
```
n8maninger, line 603: Probably prefer to remove the embed over this

ChrisSchinnerl: Yeah it's not great, but adding all the fields of the currently embedded type to the explorer types might be preferable here.

n8maninger (approved):

PR #235, opened 2025-05-21
Add file contract event tests

chris124567:
Add:
- `TestEventFileContractValid` and `TestEventFileContractMissed` for v1 file contract resolution events
- `TestEventV2FileContract` for v2 file contract resolution events

Remove old:
- `TestFileContract`
- `TestMultipleReorgFileContract`
- `TestV2FileContractResolution`

Also, refactor `scanEvents` to not use its own queries for retrieving elements and instead just use the same underlying implementation as `SiacoinElements`/`Contracts`/etc.  This requires 2-3 queries instead of one but I think is a lot cleaner because we don't have to duplicate the logic in `scanSiacoinOutput` or `scanFileContract` all over the place.

Coverage % in `persist/sqlite` is the same before and after the changes.


persist/sqlite/v2consensus_refactored_test.go
```diff
@@ -115,6 +115,40 @@ func (n *testChain) assertV2TransactionContracts(t *testing.T, txnID types.Trans
 	}
 }
 
+// assertV2TransactionContracts asserts that the enhanced
```
peterjan, line 118:
```suggestion
// assertV2TransactionResolutions asserts that the enhanced
```

persist/sqlite/addresses.go
```diff
@@ -154,33 +154,40 @@ func (s *Store) UnspentSiafundOutputs(address types.Address, offset, limit uint6
...
 }
 
-// SiacoinElements implements explorer.Store.
-func (s *Store) SiacoinElements(ids []types.SiacoinOutputID) (result []explorer.SiacoinOutput, err error) {
-	err = s.transaction(func(tx *txn) error {
-		var encoded []any
-		for _, id := range ids {
-			encoded = append(encoded, encode(id))
-		}
+func getSiacoinElements(tx *txn, merkle bool, ids []types.SiacoinOutputID) (result []explorer.SiacoinOutput, err error) {
```
peterjan, line 157: `nit` `func getSiacoinElements(tx *txn, ids []types.SiacoinOutputID, includeProof bool) (result []explorer.SiacoinOutput, err error) {`

persist/sqlite/events.go
```diff
@@ -68,11 +68,6 @@ func scanEvent(tx *txn, s scanner) (ev explorer.Event, eventID int64, err error)
 
 	switch ev.Type {
 	case wallet.EventTypeV1Transaction:
-		var txnID int64
-		err = tx.QueryRow(`SELECT transaction_id FROM v1_transaction_events WHERE event_id = ?`, eventID).Scan(&txnID)
-		if err != nil {
-			return explorer.Event{}, 0, fmt.Errorf("failed to fetch v1 transaction ID: %w", err)
-		}
 		txns, err := getTransactions(tx, []types.TransactionID{types.TransactionID(ev.ID)})
 		if err != nil || len(txns) == 0 {
```
peterjan, line 72: seems like this should return a not found error

peterjan (approved):

PR #238, opened 2025-05-23
Finish refactoring consensus tests

chris124567:
- Add
    - `TestMetricsTotalHosts`, test the TotalHosts count returned by Metrics when adding hosts.  Ensure that hosts with same pubkey aren't counted twice.
    - `TestHostScan`, where we add announce a host and then check that the counts are updated correctly when successful and unsuccessful scan happen
    - `TestV2HostScan`, which does the same thing as `TestHostScan` but for a v2 host
    - `TestV2ArbitraryData`, which tests that we can retrieve the arbitrary data field on a v2 transaction
    - `TestV2MinerFee`, which tests that we can retrieve the miner fee field on a v2 transaction
    - `TestV2FoundationAddress`, which tests that we can retrieve the new foundation address field on a v2 transaction
    - `TestV2Attestations`, which tests that we can retrieve the attestations on a v2 transaction
    - Renew the renewal contract and associated checks in `TestV2FileContractRenewal` so we don't lose any coverage from removing `TestV2FileContractRenewedToFrom`
- Remove:
    - Old test helpers
    - Old tests
        - `TestHostAnnouncement`
        - `TestV2ArbitraryData`
        - `TestV2MinerFee`
        - `TestV2FoundationAddress`
        - `TestV2Attestations`
        - `TestV2FileContractRenewedToFrom`


peterjan (approved):

chris124567:
The only thing I did was rename.  I guess I'll just do that in a separate PR.

Issue #201, opened 2025-04-09
Benchmark important queries


chris124567:
Add benchmarks for retrieving transactions, contracts, unspent outputs, etc.

chris124567:
Should be OK to close?

https://github.com/SiaFoundation/explored/pull/251

https://github.com/SiaFoundation/explored/pull/243

https://github.com/SiaFoundation/explored/pull/246

PR #243, opened 2025-05-27
Add basic benchmarks for outputs, unspent outputs, and transactions

chris124567:
Latest benchmarks:
```
BenchmarkTransactions/10_transactions
BenchmarkTransactions/10_transactions-4                      940           1219615 ns/op
BenchmarkTransactions/100_transactions
BenchmarkTransactions/100_transactions-4                     142           8281356 ns/op
BenchmarkTransactions/1000_transactions
BenchmarkTransactions/1000_transactions-4                     14          78907892 ns/op

BenchmarkSiacoinOutputs/10_unspent_outputs
BenchmarkSiacoinOutputs/10_unspent_outputs-4                6830            230802 ns/op
BenchmarkSiacoinOutputs/100_unspent_outputs
BenchmarkSiacoinOutputs/100_unspent_outputs-4               1513            862293 ns/op
BenchmarkSiacoinOutputs/1000_unspent_outputs
BenchmarkSiacoinOutputs/1000_unspent_outputs-4               164           7006526 ns/op

BenchmarkSiacoinOutputs/10_siacoin_elements
BenchmarkSiacoinOutputs/10_siacoin_elements-4               4574            272862 ns/op
BenchmarkSiacoinOutputs/100_siacoin_elements
BenchmarkSiacoinOutputs/100_siacoin_elements-4               968           1285014 ns/op
BenchmarkSiacoinOutputs/1000_siacoin_elements
BenchmarkSiacoinOutputs/1000_siacoin_elements-4               96          12238006 ns/op
```
https://github.com/SiaFoundation/explored/issues/201

### DIFF
--- a/persist/sqlite/consensus_refactored_test.go
+++ b/persist/sqlite/consensus_refactored_test.go
@@ -2666,52 +2666,11 @@ func BenchmarkTransactions(b *testing.B) {
 }
 
 func BenchmarkSiacoinOutputs(b *testing.B) {
-	pk1 := types.GeneratePrivateKey()
-	uc1 := types.StandardUnlockConditions(pk1.PublicKey())
-	addr1 := uc1.UnlockHash()
-
-	pk2 := types.GeneratePrivateKey()
-	uc2 := types.StandardUnlockConditions(pk2.PublicKey())
-	addr2 := uc2.UnlockHash()
-
-	pk3 := types.GeneratePrivateKey()
-	uc3 := types.StandardUnlockConditions(pk3.PublicKey())
-	addr3 := uc3.UnlockHash()
-
-	n := newTestChain(b, false, func(network *consensus.Network, genesisBlock types.Block) {
-		genesisBlock.Transactions[0].SiacoinOutputs[0].Address = addr1
-	})
-	genesisTxn := n.genesis().Transactions[0]
-
-	txn1 := types.Transaction{
-		SiacoinInputs: []types.SiacoinInput{{
-			ParentID:         genesisTxn.SiacoinOutputID(0),
-			UnlockConditions: uc1,
-		}},
-		SiacoinOutputs: func() (result []types.SiacoinOutput) {
-			val := genesisTxn.SiacoinOutputs[0].Value
-			// send 1000 small outputs to addr2
-			for range 1000 {
-				amount := types.NewCurrency64(1)
-				result = append(result, types.SiacoinOutput{
-					Address: addr2,
-					Value:   amount,
-				})
-				val = val.Sub(amount)
-			}
-			// send change to addr3
-			result = append(result, types.SiacoinOutput{
-				Address: addr3,
-				Value:   val,
-			})
-			return
-		}(),
-	}
-	testutil.SignTransaction(n.tipState(), pk1, &txn1)
-
-	n.mineTransactions(b, txn1)
+	addr1 := types.StandardUnlockConditions(types.GeneratePrivateKey().PublicKey()).UnlockHash()
+	n := newTestChain(b, false, nil)
 
 	// add a bunch of random outputs
+	var ids []types.SiacoinOutputID
 	err := n.db.transaction(func(tx *txn) error {
 		stmt, err := tx.Prepare(`INSERT INTO siacoin_elements(block_id, output_id, leaf_index, spent_index, source, maturity_height, address, value) VALUES (?, ?, ?, ?, ?, ?, ?, ?)`)
 		if err != nil {
@@ -2721,15 +2680,18 @@ func BenchmarkSiacoinOutputs(b *testing.B) {
 		spentIndex := encode(n.tipState().Index)
 		bid := encode(n.tipState().Index.ID)
 		val := encode(types.NewCurrency64(1))
+
 		var addr types.Address
-		for i := range 5_000_000 {
+		for i := range 1_000_000 {
 			if i%100_000 == 0 {
 				b.Log("Inserted siacoin element:", i)
 			}
-			var oid types.SiacoinOutputID
-			frand.Read(oid[:])
 
-			// label half of inputs spent
+			var scID types.SiacoinOutputID
+			frand.Read(scID[:])
+			ids = append(ids, scID)
+
+			// label half of elements spent
 			var spent any
 			if i%2 == 0 {
 				spent = spentIndex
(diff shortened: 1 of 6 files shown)


persist/sqlite/consensus_refactored_test.go
```diff
@@ -2655,6 +2766,43 @@ func BenchmarkSiacoinOutputs(b *testing.B) {
 
 	n.mineTransactions(b, txn1)
 
+	// add a bunch of random outputs
```
peterjan, line 2769: similar comment here, you should call `db.UnspentSiacoinOutputs(bm.addr, bm.offset, 100)` with a random offset and limit that you know is within bounds, so you can assert you always get `limit` number of unspent outputs

for `db.SiacoinElements` you have to take a similar approach where we test it with limits 10-100-1000 rather than always fetching a single element

n8maninger:
> I suspect the bulk of the difference between the unspent outputs and siacoin elements queries is that we retrieve the merkle proof in `SiacoinElements` which can require multiple queries. Not sure if we want to do anything about it, we could try going back to a separate on disk representation of the proofs instead of using sqlite.

The problem isn't the proof storage, it's that the retrieval is inefficient. batching + prepared statement + caching provides a significant speed up:

```
diff --git a/persist/sqlite/addresses.go b/persist/sqlite/addresses.go
index 8f2e3e7..0cee8c9 100644
--- a/persist/sqlite/addresses.go
+++ b/persist/sqlite/addresses.go
@@ -4,6 +4,7 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	"math/bits"
 
 	"go.sia.tech/core/types"
@@ -154,6 +155,50 @@ func (s *Store) UnspentSiafundOutputs(address types.Address, offset, limit uint6
 	return
 }
 
+func fillElementProofs(tx *txn, indices []uint64) (proofs [][]types.Hash256, _ error) {
+	if len(indices) == 0 {
+		return nil, nil
+	}
+
+	var numLeaves uint64
+	if err := tx.QueryRow(`SELECT num_leaves FROM network_metrics ORDER BY height DESC LIMIT 1`).Scan(decode(&numLeaves)); err != nil {
+		return nil, fmt.Errorf("failed to query state tree leaves: %w", err)
+	}
+
+	stmt, err := tx.Prepare(`SELECT value FROM state_tree WHERE row=? AND column=?`)
+	if err != nil {
+		return nil, fmt.Errorf("failed to prepare statement: %w", err)
+	}
+	defer stmt.Close()
+
+	data := make(map[uint64]map[uint64]types.Hash256)
+	for _, leafIndex := range indices {
+		proof := make([]types.Hash256, bits.Len64(leafIndex^numLeaves)-1)
+		for j := range proof {
+			row, col := uint64(j), (leafIndex>>j)^1
+
+			// check if the hash is already in the cache
+			if h, ok := data[row][col]; ok {
+				proof[j] = h
+				continue
+			}
+
+			// query the hash from the database
+			if err := stmt.QueryRow(row, col).Scan(decode(&proof[j])); err != nil {
+				return nil, fmt.Errorf("failed to query state element (%d,%d): %w", row, col, err)
+			}
+
+			// cache the hash
+			if _, ok := data[row]; !ok {
+				data[row] = make(map[uint64]types.Hash256)
+			}
+			data[row][col] = proof[j]
+		}
+		proofs = append(proofs, proof)
+	}
+	return
+}
+
 func getSiacoinElements(tx *txn, ids []types.SiacoinOutputID, includeProof bool) (result []explorer.SiacoinOutput, err error) {
 	var encoded []any
 	for _, id := range ids {
@@ -166,20 +211,27 @@ func getSiacoinElements(tx *txn, ids []types.SiacoinOutputID, includeProof bool)
 	}
 	defer rows.Close()
 
+	var leafIndices []uint64
 	for rows.Next() {
 		sco, err := scanSiacoinOutput(rows)
 		if err != nil {
 			return nil, fmt.Errorf("failed to scan siacoin output: %w", err)
 		}
-		if includeProof {
-			sco.StateElement.MerkleProof, err = merkleProof(tx, sco.StateElement.LeafIndex)
-			if err != nil {
-				return nil, fmt.Errorf("failed to get output merkle proof: %w", err)
-			}
-		}
 
+		leafIndices = append(leafIndices, sco.StateElement.LeafIndex)
 		result = append(result, sco)
 	}
+
+	if includeProof {
+		proofs, err := fillElementProofs(tx, leafIndices)
+		if err != nil {
+			return nil, fmt.Errorf("failed to fill siacoin output proofs: %w", err)
+		}
+		for i := range result {
+			result[i].StateElement.MerkleProof = proofs[i]
+		}
+	}
+
 	return
 }
 ```

```
**Before**
BenchmarkSiacoinOutputs/10_unspent_outputs-16         	   23709	     43972 ns/op	   21811 B/op	     324 allocs/op
BenchmarkSiacoinOutputs/100_unspent_outputs-16        	    7676	    163466 ns/op	  139756 B/op	    2127 allocs/op
BenchmarkSiacoinOutputs/1000_unspent_outputs-16       	     844	   1344622 ns/op	 1225738 B/op	   20132 allocs/op
BenchmarkSiacoinOutputs/10_siacoin_elements-16        	    9141	    129601 ns/op	   64923 B/op	    1072 allocs/op
BenchmarkSiacoinOutputs/100_siacoin_elements-16       	     753	   1571962 ns/op	  585470 B/op	    9593 allocs/op
BenchmarkSiacoinOutputs/1000_siacoin_elements-16      	      67	  15921842 ns/op	 5679791 B/op	   94665 allocs/op

**After**
BenchmarkSiacoinOutputs/10_unspent_outputs-16         	   23065	     51646 ns/op	   21810 B/op	     324 allocs/op
BenchmarkSiacoinOutputs/100_unspent_outputs-16        	    7718	    163920 ns/op	  139752 B/op	    2127 allocs/op
BenchmarkSiacoinOutputs/1000_unspent_outputs-16       	     880	   1316732 ns/op	 1225736 B/op	   20132 allocs/op
BenchmarkSiacoinOutputs/10_siacoin_elements-16        	   22579	     52932 ns/op	   40811 B/op	     486 allocs/op
BenchmarkSiacoinOutputs/100_siacoin_elements-16       	    4558	    264634 ns/op	  313874 B/op	    3052 allocs/op
BenchmarkSiacoinOutputs/1000_siacoin_elements-16      	     452	   2621011 ns/op	 2928436 B/op	   28539 allocs/op
```

It's still slower but acceptable. The initial 15ms is pretty acceptable.

PR #246, opened 2025-06-04
Speed up reverts subtantially

chris124567:
- Add indices for all foreign key columns
- Remove ON CASCADE DELETE and handle deletion manually in order
- Add migration (unfortunately, SQLite does not allow adjusting whether a column cascade deletes, so it requires recreating tables and copying data)

Reverting a block previously took several minutes but now takes between tens and hundreds of ms.


persist/sqlite/migrations.go
```diff
@@ -1,6 +1,10 @@
 package sqlite
 
-import "go.uber.org/zap"
+import (
```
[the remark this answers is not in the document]
peterjan: It's only used in `TestMigrationConsistency`

persist/sqlite/consensus_refactored_test.go
```diff
@@ -2814,3 +2816,96 @@ func BenchmarkSiacoinOutputs(b *testing.B) {
...
+
+	ru := consensus.RevertBlock(prevState, block, bs)
+	crus := []chain.RevertUpdate{{
+		RevertUpdate: ru,
+		Block:        block,
+		State:        prevState,
+	}}
+
+	b.ResetTimer()
+	for range b.N {
```
peterjan, line 2896: `nit` use `b.Loop()`

chris124567:
ok performance without temp tables is slightly better 
```
BenchmarkRevert-4             43          26088574 ns/op
```

persist/sqlite/consensus_refactored_test.go
chris124567: I tried this in the first benchmark PR and got a CI error.  I think it's a very recent addition to Go

persist/sqlite/revert.go
```diff
@@ -2,53 +2,242 @@ package sqlite
...
-		FROM block_transactions bt
-		WHERE bt.transaction_id = t.id
-);`)
+func deleteV1Transactions(tx *txn, bid types.BlockID) error {
+	rows, err := tx.Query(`SELECT transaction_id AS id
+FROM
+	block_transactions
+WHERE
+	block_id = ?
+	AND transaction_id NOT IN (
```
peterjan, line 78: I wonder if it's noticeably faster to `GROUP BY` transaction_id and have a distint count on block_id that equals 1. So something like

```sql
SELECT transaction_id
FROM block_transactions
GROUP BY transaction_id
HAVING COUNT(DISTINCT block_id) = 1 AND ANY(block_id) = ?
```

peterjan: Thanks for checking

peterjan (approved):

chris124567:
```
BenchmarkRevert-4             45          26976586 ns/op
```

peterjan (dismissed):

ChrisSchinnerl (dismissed):

persist/sqlite/migrations.go
ChrisSchinnerl, line 1: Blocker to not merge this until siascan is synced and working again.

PR #247, opened 2025-06-05
Fix case where some payout events were not added

chris124567:
`AppliedEvents` would return early if it encountered a file contract element diff that was not a resolution instead of continuing the loop, which would result in miner payout events and foundation subsidy events not being in included in the list of events if a block contained contract formations or revisions.  This fixes that and adds a related test.

### DIFF
--- a/explorer/events.go
+++ b/explorer/events.go
@@ -465,7 +465,7 @@ func AppliedEvents(cs consensus.State, b types.Block, cu ChainUpdate) (events []
 	for _, diff := range cu.FileContractElementDiffs() {
 		fce, resolved, valid := diff.FileContractElement, diff.Resolved, diff.Valid
 		if !resolved {
-			return
+			continue
 		}
 
 		fce.StateElement.MerkleProof = nil
@@ -518,7 +518,7 @@ func AppliedEvents(cs consensus.State, b types.Block, cu ChainUpdate) (events []
 	for _, diff := range cu.V2FileContractElementDiffs() {
 		fce, res := diff.V2FileContractElement, diff.Resolution
 		if res == nil {
-			return
+			continue
 		}
 
 		fce.StateElement.MerkleProof = nil
--- a/persist/sqlite/consensus_refactored_test.go
+++ b/persist/sqlite/consensus_refactored_test.go
@@ -2558,6 +2558,70 @@ func TestHostScan(t *testing.T) {
 	assertHost(hosts[0].PublicKey, host1)
 }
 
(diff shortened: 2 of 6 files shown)


persist/sqlite/consensus_refactored_test.go
```diff
@@ -2558,6 +2558,70 @@ func TestHostScan(t *testing.T) {
...
+	addr2 := uc2.UnlockHash()
+
+	n := newTestChain(t, false, func(network *consensus.Network, genesisBlock types.Block) {
+		genesisBlock.Transactions[0].SiacoinOutputs[0].Address = addr1
+	})
+	genesisTxn := n.genesis().Transactions[0]
+
+	fc := prepareContract(addr1, n.tipState().Index.Height+1)
+	// create file contract
+	txn1 := types.Transaction{
```
ChrisSchinnerl, line 2580: could you also add a v2 txn to make sure the second occurrence gets hit as well?

explorer/events.go
ChrisSchinnerl, line 1: We are also going to need a migration. The question is whether there is a nice way to fix only the missing payouts and subsidy or we just write a migration that performs a full resync. Kubernetes should make it easy to kill one pod and sync and then kill the other one without downtime. Also if we need a full resync anyway, the other PR's migration might become simpler as well.

@n8maninger any preference?

chris124567: do we have a backup instance running we can point siascan.com to even if the payout events are missing? full reset will take a day or two to run which might not be desirable given hardfork

ChrisSchinnerl: Right now we load balance siascan between 2 instances of explored. So we can just resync one at a time. 
Probably best to discuss that on Slack.

persist/sqlite/consensus_refactored_test.go
```diff
@@ -2558,6 +2558,70 @@ func TestHostScan(t *testing.T) {
 	assertHost(hosts[0].PublicKey, host1)
 }
 
+func TestEventPayoutContract(t *testing.T) {
```
peterjan, line 2561: `nit` instead of leaving a comment inside of the test, it's probably a bit cleaner to have a docstring that looks like  `// TestEventPayoutContract is a regression test that covers a bug in ...`

ChrisSchinnerl (commented):
@chris124567 looks good overall. Before we merge this, did you test the migration manually to make sure it's working?

ChrisSchinnerl (approved):

PR #256, opened 2025-06-26
Split tests into multiple files

chris124567:
Move tests from `consensus_refactored_test.go` and `v2consensus_refactored_test.go` into more specific files like `transactions_test.go` or `metrics_test.go`


peterjan:
Cherry on top could be extracting `testChain` into a separate file. That's probably close to 800 lines by itself.

ChrisSchinnerl (approved):

Issue #208, opened 2025-04-23
Clean up database code


ChrisSchinnerl:
Now that we are mostly feature complete it's time to revisit the database code. There are various places where we can move code around to clean things up. e.g. returning database IDs or even maps of them might be a sign that we could improve these methods by moving the code that needs the IDs closer to the the one reading them. Ideally IDs stay within the scope of the function using them rather than being passed around too much. We might also be able to make use of upserts in some places.

chris124567:
Was this completed by #218 or were there other sections of the code you had in mind?  There are probably some modest improvements that could be made in `persist/sqlite/consensus.go` but I don't see anything radically different being possible unless we have a different schema.

chris124567:
After #258, the only maps of IDs we have are for transactions.  We could probably get rid of these if we wanted, but I think it might make things worse:

```
func addSignatures(tx *txn, id int64, txn types.Transaction) error {
    stmt, err := tx.Prepare(`INSERT INTO transaction_signatures(transaction_id, transaction_order, parent_id, public_key_index, timelock, covered_fields, signature) VALUES (?, ?, ?, ?, ?, ?, ?)`)
    if err != nil {
        return fmt.Errorf("addMinerFees: failed to prepare statement: %w", err)
    }
    defer stmt.Close()

    for i, sig := range txn.Signatures {
        if _, err := stmt.Exec(id, i, encode(sig.ParentID), sig.PublicKeyIndex, encode(sig.Timelock), encode(sig.CoveredFields), sig.Signature); err != nil {
            return fmt.Errorf("addMinerFees: failed to execute statement: %w", err)
        }
    }
    return nil
}
```

vs:

```
func addSignatures(tx *txn, txn types.Transaction) error {
    stmt, err := tx.Prepare(`INSERT INTO transaction_signatures(transaction_id, transaction_order, parent_id, public_key_index, timelock, covered_fields, signature) VALUES ((SELECT id FROM transactions WHERE transaction_id = ?), ?, ?, ?, ?, ?, ?)`)
    if err != nil {
        return fmt.Errorf("addMinerFees: failed to prepare statement: %w", err)
    }
    defer stmt.Close()

    for i, sig := range txn.Signatures {
        if _, err := stmt.Exec(encode(txn.ID()), i, encode(sig.ParentID), sig.PublicKeyIndex, encode(sig.Timelock), encode(sig.CoveredFields), sig.Signature); err != nil {
[...]

PR #258, opened 2025-06-30
Clean up inserts

chris124567:
Remove usage of maps that map v1/v2 file contract element IDs, siacoin element IDs, and siafund elements IDs to their integer IDs in the database and instead retrieve those IDs using subqueries.

`master`
```
BenchmarkApplyRevert/apply-4                 246       4828469 ns/op
BenchmarkApplyRevert/revert-4                 33      33822119 ns/op
```

`christopher/clean-up-inserts`:
```
BenchmarkApplyRevert/apply-4                 247       5129194 ns/op
BenchmarkApplyRevert/revert-4                 34      33314049 ns/op
```

Given that indexing is already pretty slow I'm not sure whether we are OK with a slowdown though this one is relatively small.

https://github.com/SiaFoundation/explored/issues/208

### DIFF
--- a/persist/sqlite/consensus_test.go
+++ b/persist/sqlite/consensus_test.go
@@ -813,7 +813,9 @@ func BenchmarkAddressEvents(b *testing.B) {
 	}
 }
 
-func BenchmarkRevert(b *testing.B) {
+func BenchmarkApplyRevert(b *testing.B) {
+	const nBlocks = 20_000
+
 	pk1 := types.GeneratePrivateKey()
 	uc1 := types.StandardUnlockConditions(pk1.PublicKey())
 	addr1 := uc1.UnlockHash()
@@ -829,11 +831,7 @@ func BenchmarkRevert(b *testing.B) {
 	val := genesisTxn.SiacoinOutputs[0].Value
 	scID := genesisTxn.SiacoinOutputID(0)
 
-	for i := range 20000 {
-		if i%100 == 0 {
-			b.Logf("Mined %d blocks", i)
-		}
-
+	generateBlock := func() types.Block {
 		// transaction with random arbitrary data
 		data := make([]byte, 16)
 		frand.Read(data)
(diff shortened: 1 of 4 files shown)


persist/sqlite/consensus_test.go
```diff
@@ -882,12 +880,80 @@ func BenchmarkRevert(b *testing.B) {
...
+	cs, au := consensus.ApplyBlock(cs, block, bs, time.Time{})
+	caus := []chain.ApplyUpdate{{
+		ApplyUpdate: au,
+		Block:       block,
+		State:       cs,
+	}}
+
+	b.Run("apply", func(b *testing.B) {
+		b.ResetTimer()
+		for range b.N {
```
peterjan, line 933: `nit` use `b.Loop`

persist/sqlite/consensus_test.go
```diff
@@ -882,12 +880,80 @@ func BenchmarkRevert(b *testing.B) {
...
+			err := n.db.transaction(func(tx *txn) error {
+				defer tx.Rollback()
```
peterjan, line 935: I think this is fine but you could also have the function return a custom error, which you can then ignore. If that function returns an error `db.transaction` will roll back for you. I would probably `b.Fatal` if `UpdateChainState` fails too simply because it's unexpected.

persist/sqlite/consensus_test.go
```diff
@@ -882,12 +880,80 @@ func BenchmarkRevert(b *testing.B) {
...
+				utx := &updateTx{
+					tx: tx,
+				}
+
+				b.StartTimer()
```
peterjan, line 940: Looks slightly weird to me because it's usually the inverse, but it's probably fine though. I would replace `b.ResetTimer` with `b.StopTimer`, because resetting is redundant and if you stop it timer it's even more correct.

persist/sqlite/consensus_test.go
```diff
@@ -882,12 +880,80 @@ func BenchmarkRevert(b *testing.B) {
...
+				err := explorer.UpdateChainState(utx, nil, caus, n.db.log.Named("update"))
+				b.StopTimer()
+
+				if err != nil {
+					return fmt.Errorf("failed to update chain state: %w", err)
+				}
+				return nil
+			})
+			if err != nil && !strings.Contains(err.Error(), "rolled back") {
```
peterjan, line 949: Instead of using a transaction that you can roll back, did you consider reverting the block outside the timer and vice versa in the revert? That way you isolate the benchmark around `UpdateChainState` and you have something like:

```go
for b.Loop() {
    if err := explorer.UpdateChainState(utx, nil, caus, n.db.log.Named("update")); err != nil { b.Fatal(err) }
    b.StopTimer()
    if err := explorer.UpdateChainState(utx, crus, nil, n.db.log.Named("update")); err != nil { b.Fatal(err) } // revert
    b.StartTimer()
}
```

I thought rolling back the transaction was pretty clever actually, but then I realised the way you set it up you never actually commit the transaction because you manually rolled it back before that point. I think that's probably illegal in a benchmark, or at the very least not the best practice. We definitely want the commit within the bounds of what we're benchmarking.

chris124567:
`master`:
```
BenchmarkApplyRevert/apply-4                 254       4762309 ns/op
BenchmarkApplyRevert/revert-4                 32      34342211 ns/op
```


`christopher/clean-up-inserts`:
```
BenchmarkApplyRevert/apply-4                 243       4839120 ns/op
BenchmarkApplyRevert/revert-4                 33      33778788 ns/op
```

PR #264, opened 2025-07-15
Be more consistent with error messages

chris124567:
Many helper functions like those used to add transaction fields or update balances are only called in one or two places.  Instead of littering the function's name throughout the error messages in the function, instead just use the name of the function in the error message at the callsite.  This is already done in some places, but inconsistently.

Example:
```
 func addMinerPayouts(tx *txn, bid types.BlockID, scos []types.SiacoinOutput) error {
        stmt, err := tx.Prepare(`INSERT INTO miner_payouts(block_id, block_order, output_id) VALUES (?, ?, (SELECT id FROM siacoin_elements WHERE output_id = ?));`)
        if err != nil {
-               return fmt.Errorf("addMinerPayouts: failed to prepare statement: %w", err)
+               return fmt.Errorf("failed to prepare statement: %w", err)
        }
        defer stmt.Close()
 
        for i := range scos {
                if _, err := stmt.Exec(encode(bid), i, encode(bid.MinerOutputID(i))); err != nil {
-                       return fmt.Errorf("addMinerPayouts: failed to execute statement: %w", err)
+                       return fmt.Errorf("failed to execute statement: %w", err)
                }
        }
```

... later in `ApplyIndex`

```
        } else if err := addMinerPayouts(ut.tx, state.Block.ID(), state.Block.MinerPayouts); err != nil {
-               return fmt.Errorf("ApplyIndex: failed to add miner payouts: %w", err)
+               return fmt.Errorf("failed to add miner payouts: addMinerPayouts: %w", err)
```

### DIFF
--- a/explorer/update.go
+++ b/explorer/update.go
@@ -470,10 +470,7 @@ func revertChainUpdate(tx UpdateTx, cru chain.RevertUpdate, revertedIndex types.
 	}
 	state.Metrics.Index = revertedIndex
 
-	if err := tx.RevertIndex(state); err != nil {
-		return fmt.Errorf("failed to revert block: %w", err)
-	}
-	return nil
+	return tx.RevertIndex(state)
 }
 
 func updateMetrics(tx UpdateTx, s UpdateState, metrics Metrics) (Metrics, error) {
--- a/explorer/update.go
+++ b/explorer/update.go
@@ -325,7 +325,10 @@ func applyChainUpdate(tx UpdateTx, cau chain.ApplyUpdate) error {
 	state.Metrics.SiafundTaxRevenue = cau.State.SiafundTaxRevenue
 	state.Metrics.NumLeaves = cau.State.Elements.NumLeaves
 
-	return tx.ApplyIndex(state)
+	if err := tx.ApplyIndex(state); err != nil {
+		return fmt.Errorf("failed to apply block: ApplyIndex: %w", err)
+	}
+	return nil
 }
 
 // revertChainUpdate atomically reverts a chain update from a store
@@ -467,7 +470,10 @@ func revertChainUpdate(tx UpdateTx, cru chain.RevertUpdate, revertedIndex types.
 	}
 	state.Metrics.Index = revertedIndex
 
-	return tx.RevertIndex(state)
+	if err := tx.RevertIndex(state); err != nil {
+		return fmt.Errorf("failed to revert block: RevertIndex: %w", err)
+	}
+	return nil
 }
 
 func updateMetrics(tx UpdateTx, s UpdateState, metrics Metrics) (Metrics, error) {
--- a/explorer/update.go
+++ b/explorer/update.go
@@ -318,7 +318,7 @@ func applyChainUpdate(tx UpdateTx, cau chain.ApplyUpdate) error {
 	}
 	state.Metrics, err = updateMetrics(tx, state, prevMetrics)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to update metrics: %w", err)
 	}
 	state.Metrics.Index = cau.State.Index
 	state.Metrics.Difficulty = cau.State.Difficulty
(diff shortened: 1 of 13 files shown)


explorer/update.go
```diff
@@ -467,7 +470,10 @@ func revertChainUpdate(tx UpdateTx, cru chain.RevertUpdate, revertedIndex types.
 	}
 	state.Metrics.Index = revertedIndex
 
-	return tx.RevertIndex(state)
+	if err := tx.RevertIndex(state); err != nil {
```
n8maninger, line 473: Should be fine to punt this up the stack since there's only one possible return site

explorer/update.go
```diff
@@ -325,7 +325,10 @@ func applyChainUpdate(tx UpdateTx, cau chain.ApplyUpdate) error {
 	state.Metrics.SiafundTaxRevenue = cau.State.SiafundTaxRevenue
 	state.Metrics.NumLeaves = cau.State.Elements.NumLeaves
 
-	return tx.ApplyIndex(state)
+	if err := tx.ApplyIndex(state); err != nil {
+		return fmt.Errorf("failed to apply block: ApplyIndex: %w", err)
```
n8maninger, line 329: can drop the `ApplyIndex`. Should wrap the error on :321

persist/sqlite/consensus.go
```diff
@@ -185,13 +185,13 @@ func addFileContracts(tx *txn, dbID int64, txn types.Transaction) error {
...
 	if err != nil {
-		return fmt.Errorf("addFileContracts: failed to prepare statement: %w", err)
+		return fmt.Errorf("failed to prepare statement: %w", err)
 	}
 	defer stmt.Close()
 
 	for i, fc := range txn.FileContracts {
 		if _, err := stmt.Exec(dbID, i, encode(txn.FileContractID(i)), encode(fc.RevisionNumber)); err != nil {
-			return fmt.Errorf("addFileContracts: failed to execute transaction_file_contracts statement: %w", err)
+			return fmt.Errorf("failed to execute transaction_file_contracts statement: %w", err)
```
n8maninger, line 194: `failed to execute statement:` is fine here too

persist/sqlite/consensus.go
```diff
@@ -325,25 +325,25 @@ func addTransactionFields(tx *txn, txns []types.Transaction, txnExist map[types.
 		}
 
 		if err := addMinerFees(tx, dbID, txn.MinerFees); err != nil {
-			return fmt.Errorf("failed to add miner fees: %w", err)
+			return fmt.Errorf("failed to add miner fees: addMinerFees: %w", err)
```
n8maninger, line 328: `addMinerFees:` is unnecessary. `failed to add miner fees:` is unique enough context to track it down. Ditto for all of these.

persist/sqlite/consensus.go
```diff
@@ -1008,19 +1008,19 @@ func (ut *updateTx) Metrics(height uint64) (explorer.Metrics, error) {
 
 func (ut *updateTx) ApplyIndex(state explorer.UpdateState) error {
 	if err := addBlock(ut.tx, state.Block, state.ChainIndexElement, state.Metrics.Index.Height); err != nil {
-		return fmt.Errorf("ApplyIndex: failed to add block: %w", err)
+		return fmt.Errorf("failed to add block: addBlock: %w", err)
```
n8maninger, line 1011: `addBlock:` is unnecessary. Same with the rest

persist/sqlite/consensus.go
```diff
@@ -1227,15 +1227,16 @@ func (s *Store) ResetChainState() error {
 	if err := s.transaction(func(tx *txn) error {
 		if _, err := tx.Exec(`PRAGMA defer_foreign_keys=ON`); err != nil {
 			return fmt.Errorf("failed to defer foreign key checks: %w", err)
+		} else if err := resetChainState(tx, s.log, int64(len(migrations)+1)); err != nil {
+			return fmt.Errorf("failed to reset chain state: resetChainState: %w", err)
```
n8maninger, line 1231: extra context `resetChainState:`

n8maninger (changes requested):
I didn't get through every file, but the goal with adding context is to provide just enough info to identify the error site through the stack without duplication. There's a lot that could still be trimmed.

explorer/update.go
n8maninger: @chris124567 doesn't look like this was changed.

chris124567: Should be fixed now

PR #267, opened 2025-08-12
api: Add /metrics/difficulty

lukechampine:
Adds a `/metrics/difficulty` endpoint:
```
GET /metrics/difficulty?start=0&end=500000

Response 200:
{
  "blocksPerStep": 3333,
  "difficulties":["115312521", "119515127", "110032522", ...],
  "blockTimes": [618247, 291533, 1100934, ... ]
}
```
(Block times are given in milliseconds.)

Given a range of block heights, we determine an appropriate `step` such that we return at most 150 data points. (Picked somewhat arbitrarily; we should try other values.) This should allow the frontend to present graphs that can be panned and zoomed without having to transfer the whole data set.

As a future enhancement, we could also include the minimum and maximum values for each data point, and add fainter lines for them on the graph; that way, we don't hide outliers, even at low granularity.

### DIFF
--- a/persist/sqlite/metrics_test.go
+++ b/persist/sqlite/metrics_test.go
@@ -1,6 +1,7 @@
 package sqlite
 
 import (
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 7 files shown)


persist/sqlite/metrics_test.go
```diff
@@ -439,6 +440,81 @@ func TestV2Metrics(t *testing.T) {
...
+		},
+		{
+			name:  "Start equals end",
+			start: 5, end: 5, step: 1,
+		},
+		{
+			name:  "Invalid range",
+			start: 5, end: 3, step: 1,
+			err: "start height 5 cannot be greater than end height 3",
+		},
```
ChrisSchinnerl, line 471: (nit) might be good to add a test case for when the step doesn't cleanly divide the length and another one for step = 0.

ChrisSchinnerl (approved):

Issue #273, opened 2025-08-26
Add OpenAPI spec


n8maninger:
Add an OpenAPI spec similar to `renterd` and `indeed`

PR #280, opened 2025-10-08
Add OpenAPI spec

peterjan:
This PR adds an OpenAPI spec for walletd. I'm going to do the same for walletd and hostd. This is in reference to one of the [goals we set last year](https://www.notion.so/siafoundation/New-API-docs-189632e3aeaf80c1a7e8d9640a724db4#189632e3aeaf80c1a7e8d9640a724db4), but we failed to accomplish it in due time. But better late than never.

I'm not too familiar with Scalar, I copy pasted the workflow from indexd and tweaked it a bit. The openapi spec was generated of course, I verified it manually though and ran it through a validator.

Fixes https://github.com/SiaFoundation/explored/issues/273

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -27,8 +27,6 @@ tags:
     description: Transaction lookups and chain location helpers.
   - name: addresses
     description: Address balances, UTXOs, and activity.
-  - name: rankings
-    description: Leaderboards for top addresses.
   - name: events
     description: Wallet events and UTXO lookups.
   - name: contracts


.github/workflows/openapi-publish.yml
```diff
@@ -0,0 +1,21 @@
...
+      - "openapi.yml"
+  workflow_dispatch:
+
+jobs:
+  publish:
+    uses: SiaFoundation/workflows/.github/workflows/publish-openapi.yml@master
+    with:
+      slug: explored
+      spec_path: openapi.yml
+      docs_slug: sia
```
alexfreska, line 19: Set the docs_slug to `explored` since we want a separate docs site per daemon.

openapi.yml
```diff
@@ -0,0 +1,2389 @@
...
+        address:
+          $ref: "#/components/schemas/Address"
+        spent:
+          type: boolean
+        spentTransactionID:
+          allOf:
+            - $ref: "#/components/schemas/TransactionID"
+            - nullable: true
+        source:
+          type: string
```
chris124567, line 1394: add
```
enum: ["invalid", "miner_payout", "transaction", "valid_proof_output", "missed_proof_output"]
```

openapi.yml
```diff
@@ -0,0 +1,2389 @@
...
+            - nullable: true
+      additionalProperties: true
+
+    ExplorerEvent:
+      type: object
+      properties:
+        id:
+          $ref: "#/components/schemas/Hash256"
+        type:
+          type: string
```
chris124567, line 1422: add to `type`:
```
enum: ["miner", "foundation", "siafundClaim", "v1Transaction", "v1ContractResolution", "v2Transaction", "v2ContractResolution"]
```

